PR #38, opened 2015-09-09
Support Mac OS X build

tomyun:
Here is a small set of changes that could make taisei run on Mac OS X with libraries built with [Homebrew](http://brew.sh/).

- Use bundled `OpenGL.framework`
  - `OpenGL/` instead of `GL/` for headers
  - No need for importing GL functions (no `glXGetProcAddress` from `glx.h`)

Yet, it still gives me no sound though. Probably something goes wrong with `OpenAL.framework` (or `openal-soft`) and `freealut` packages.

### DIFF
--- a/src/resource/shader.h
+++ b/src/resource/shader.h
@@ -8,7 +8,11 @@
 #ifndef SHADER_H
 #define SHADER_H
 
+#ifdef __APPLE__
+#include <OpenGL/gl.h>
+#else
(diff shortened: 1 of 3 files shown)


laochailan:
Looks good, thanks!

laochailan:
I don’t know about the sound, but I remember earlier tries to build it on osx also failed at sound issues.

PR #40, opened 2016-12-19
Integral complex constants changed to float

makise-homura:
Because of some compilers don't support complex integer arithmetic (which is GCC extension), and source code of taisei contains such numbers only in constants, I decided to replace every such constant to floating-point one, as arithmetic really is floating-point in many cases (contains divisions, floating-point numbers in expressions, etc.).

This may help to achieve portability to platforms for which gcc or replacing compiler does not offer integral arithmetic option (and did helped so for E2K platform).

Build was tested and no strange behavior or errors were detected.


laochailan:
Cool, thanks!

PR #42, opened 2017-01-24
Background music support

makise-homura:
Since Touhou series isn't so perfect without music, I decided to add background music support to Taisei.
It may be in .wav or .ogg (if compiled with libogg, libvorbis, and libvorbisfile) format.
Should be placed in bgm/ subdirectory under special names (see README).
Stage music may have its titles displayed while started to play; these titles are configured in bgm/bgm.conf.

It leads to a strong sound subsystem rework, so there's adiitional changes:
- optional OGG format support for BGM and SFX;
- SFX in menu;
- in-stage BGM titles;
- Volume level setup for BGM and SFX;
- BGM switch support in dialogs (in addition to 'Left' and 'Right' phrases, there is 'BGM' now).

TODO (ideas to work on in future):
- speed up music loading (it takes about 1-2 second for each ~40MB .wav or 3-4 seconds for ~4MB ogg).
- display loading progress on start screen (as it takes much more time now with BGM enabled);
- find corresponding BGM under MIT-compartible license... or wait until any other contributor decide to do so. :)


laochailan:
Awesome work :) it is great to see somebody put some life to an old project like this. I hope the code isn’t too messy and mysterious.

I don’t have much time to give it any thorough testing, but I read your code and ran it once. It looks really good!

One minor issue: If I try to build it, it doesn’t work unless I remove the inline before detect_endianness() in ogg.c (or add static inline). As far as I remember taisei has pretty pedantic build flags which forbid non-standard GNU features and apparently inline without static won’t work if there is no external definition of the function given somewhere … at least in my setup.

I still haven’t ensnared anyone to create a soundtrack for this.

But I’ve since then much improved on my painting skills and I’m kind of motivated now to maybe paint all the missing dialog graphics and repaint the existing ones. :)

makise-homura:
Thank you!
As far as I was able to find, that's the only Touhou clone which can be built for native Linux (and non-x86 architectures, which is critical for me) and, in addition, licensed under one of free licenses. So it's pleasure for me to do some work on it. :)
(BGM feature was mainly the most wide-scale improvement to Taisei I thought of, but I have some minor ideas too, so I'll continue working on it.)

Code seems sufficiently clear to me, so it's no problem for me to understand, for example, audio-related part. (But the one which related to stage mechanics, was not so much intuitively understandable; so I preferred to note Kurumi's behavior as issue rather than commit & pull request, since I do not understand completely which side effects may arise if I change stage timing).

According detect_endianness(), my gcc 4.8.1 does not show any warnings (-Wall was specified in CMakeLists.txt), so I was not able to see this. Well, I think, it may be fixed in either of possible ways - as this function called only once for each .ogg file that is about to load, so there is no noticeable effect in inline-related optimization (.ogg decoding itself consumes much more time).
Many of those who work on widespread architectures use compiler-related constants (__BYTE_ORDER__, etc.) - which is not subject to my build environment, or even hardcode endianness in their sources (last is not the case for me, too - because I want it to be as much cross-architecture as possible).

Speaking of graphics and soundtrack - it would be great if there'll be someone who could write music for Taisei, so I'll hope for such luck (Tewi, help me! :). And also reworking graphics would be nice too (since ZUN skills also had became improved. :) And then it would really be a serious project rather than write-and-abandon thing. Completely free and open source Touhou clone that may run on native linux is something that should not be left unneeded. :)

Akaricchi:
Seeing someone mess with the options code I wrote years ago makes me a bit nostalgic ;')

That said, you could've probably used `bind_scale` somehow to get nice sliders for the sfx/bgm volumes, like those used in the gamepad settings.

makise-homura:
>That said, you could've probably used bind_scale somehow to get nice sliders for the sfx/bgm volumes, like those used in the gamepad settings.
Well, it's a good idea. I'll keep this as another thing for improvement to the next bunch of fixes. :)

PR #44, opened 2017-01-30
Additional small improvements for taisei

makise-homura:
- Volume controls changed to sliders (as Akaricchi proposed in comments to previous request);
- Added debug parameter TAISEI_IDDQD to achieve infinite lives in debug mode;
- Minor fixes: old BGM does not stop if new BGM is missing; ZLIB is required by libpng, not optional; pointer conversion generates a warning on 64-bit architecture.


laochailan:
Thanks, I’ll look at this later.

Issue #41, opened 2017-01-24
Kurumi boss timing


makise-homura:
While testing stage functionality, I found that Kurumi (in boss role) timing is a bit strange.

At first, I wasn't able to reach introduction dialog with her - she starts her attacks after nearly 4 seconds of waiting.
It seems to be solved if I change AT(400) to AT(120) in src/stages/stage4_events.c:417 (kurumi_boss_intro()), but I'm not sure that I do it right.

Additionally, last attack of her final spell card (Power Sign ~ Blow the Walls) occurs after the ending dialog starts (and I have to dodge it while first dialog line is displayed).

Reproduced two times with either with Youmu and Marisa. Here goes two replays with this issue (note: they were recorded with "--" commented at src/player.c:212 to achieve some sort of infinite lives).

[replays.zip](https://github.com/laochailan/taisei/files/726664/replays.zip)

laochailan:
Truth be told, even I don’t remember the possible implications of this. I’m just a bit surprised that we didn’t notice this back in the day.

What hat i do remember though, is that we never got replays to be 100% portable, so I’m not sure if it will create the same results for me. But I will test this in a few days.

laochailan:
The dialog indeed didn’t appear. I think changing the AT like you did should do the trick. I pushed that change.

The final attack thing seems to be one of her slaves that create the bullets when they hit the wall surviving. Normally, the functions containing danmaku logic get called with the special negative time argument EVENT_DEATH when an enemy dies or a boss attack is over.

If boss attacks create slaves, it’s their responsibility to clean up so most boss attacks actually contain something like

```
AT(EVENT_DEATH)
  killall(global.enemies);
```

But Blow the Walls didn’t have this. I added it now and haven’t been able to reproduce this anymore. Can you test?

makise-homura:
Thank you for fixes! I will test it this evening (in about four-five hours from now).

makise-homura:
Thanks for the dialog (now it works), but "Blow the walls" issue is still representable (replay is attached
[taisei_20370130_22-28-32_MSK_stg4_youmuA_easy.zip](https://github.com/laochailan/taisei/files/740196/taisei_20370130_22-28-32_MSK_stg4_youmuA_easy.zip)).

[...]

PR #45, opened 2017-02-03
Another bunch of fixes

makise-homura:
- Fixed issue #41;
- Fixed mistakes in README;
- Removed some tabs at end of lines;
- As it makes sense on stage 4, added 'boss BGM title color' field to stage info.


laochailan:
Thanks, looks good. :) Maybe I should change my settings to display those tabs.

PR #46, opened 2017-02-04
SDL2 migration (WIP)

Akaricchi:
I don't know what's gotten into me, but seeing how this project got some attention again I've decided to port it to SDL2. I believe it's more mature now than the last time we tried to use it and should fix the fullscreen problems on windows. SDL1.2 is obsolete and has been so for a while; I see no reason to keep using it.

This branch is a work-in-progress, so don't merge it yet. I've created this pull request so that you can easily track and review the changes as I push them.


Akaricchi:
Note: some keybindings may become broken if you try to use a config from the SDL1.2 version. This is because our keybindings are stored as the SDL integer constants, some of which have changed in SDL2. I'll see about changing that. The workaround is to simply rebind those for now.

Akaricchi:
Follow up to the previous comment and commit, the keybindings are stored in human readable format now. IIRC this is how I wanted them to work in the first place, but SDL1.2 didn't have a good way to convert both ways. The downside is that you have to redo *all* of your keybindings now. But at least they are reset to defaults instead of being left in some broken state.

laochailan:
Cool! :) Thanks for the work! Finally, we don’t have to recreate the context all the time.

Well, config file incompatibility shouldn’t be a worry if the standard location of the config file changes. It’s a bit strange that it would put the config in .local/share though. But I guess screenshots/replays would not belong in .config and it doesn’t look like the File path API doesn’t give much of a choice.

Akaricchi:
The API doesn't give much of a choice to the programer, but it does give some to the user, since now you can set the XDG_DATA_HOME envorinment variable to change the location. It's fairly common for games to follow this standard in my experience; at least most of the Linux ports I've played on Steam do. They also use SDL2 and probably that very same API too.

Akaricchi:
@laochailan I think it's pretty solid now. Everything seems to work as intended. I've tested it on Windows too and found no problems so far. Any objections before we merge this?

laochailan:
Doesn’t cause me any problems either so we can merge I think.

Issue #48, opened 2017-02-04
Desyncing replays, tsrand, etc.


Akaricchi:
I'm still bitter about this issue and probably going to investigate it again sometime soon. I'm not sure if this is caused by incorrect usage of the tsrand functions somewhere, but I believe it's highly probable and should be tripple-checked first. At very least, the naming of those functions should definitely be revised -- what the fuck does an ```anfrand()``` do?

If you have any thoughts about replays and/or RNG, you're more than welcome to discuss them here.

laochailan:
My guess is *n* for *normal*, *f* for *float* and *a* for *a bad abbreviation*.
Now that I had my share of statistics, I understand why aiju was so mad when we only checked the mean of those functions. I may plot a histogram of them for once :smile:

Akaricchi:
That'd be welcome, though I don't think the quality of the RNG itself would be a cause for desyncing, for as long as it generates consistent sequences on all platforms.

As for floats, while it might not fix anything I think we should at least use double instead of float consistently. It's all over the place right now. You can never tell when an implicit conversion occurs. Maybe we'll even abandon them for fixed-point arithmentic eventually, but that sounds a bit scary right now, considering the amount of rewriting involved...

Another thing is that we store some initial player state in the replay, and some of it are floats (the coordinates). The replay file structure is stupid and text-based, and the textual representation of those floats may just not be accurate enough. I think I'll look into going back to a binary format for it again, once the SDL2 branch is merged. There's a supposedly portable API for raw I/O that I could use. It always bothered me how it's text based anyway, even though virtually all of the data in there are numbers.

laochailan:
Yes of course it wouldn’t be at heart of the desync issues, but I’m curious if those are really behaving like they should. (and I have an urge to apply statistics)

Switching to double might be okay, but in the end everything is going to be implicitly cast to float by all the OpenGL calls.

Fixed point arithmetic seems like a bad idea. What would be the merit? Just to get replays cross platform? That can’t be a solution. Think about the amount of lookup tables and things you would need to get right just for a fixed point carg().
If things really start getting inaccurate, we can think about saving the absolute position or something or introducing some subtle rounding in the collision checks that is not noticeable to the player but larger than round-off error/machine differences.

I think we should do neither that nor the float to double thing before we are sure that’s what is causing it.

But switching to binary would be very nice :smile:

Akaricchi:
[...]

PR #50, opened 2017-02-05
Replays II: The Movie: The Game

Akaricchi:
See also #48 and the commit messages for more information. 

Not quite ready for merge yet. Most importantly, it needs more testing, and I'll appreciate any help I can get with that.

**If you want to help me test this**, first you have to build from this branch with the following flags:

```
cmake -DTAISEI_DEBUG=1 -DTAISEI_IDDQD=1 ..
```

Then, download [this replay](https://www.dropbox.com/sh/1jkiho7tb6v9wt3/AAByK5a-QuxQZ7S8UcSsyGn8a/taisei_20170206_01-10-18_EET_youmuB_lunatic.tsr?dl=1) and place it in ```~/.local/share/taisei/replays```. It was recorded on hyper speed with god mode, so just in case you notice that my actions there don't make sense, they aren't supposed to.

Then, execute taisei like this:

```
TAISEI_SANIC=-1 taisei &> /tmp/taisei_log
```

Then watch the replay from start to finish. You won't be able to actually see anything, because the game's FPS limiter will be disabled and nothing will be rendered to let the game logic run as fast as possible.

After that, quit Taisei and upload ```/tmp/taisei_log``` to a pastebin of some kind. Then run it again, but this time like this:

```
TAISEI_SANIC=1 taisei &> /tmp/taisei_log
```

Watch replay, upload log. This time it'll be slower, because the game will actually render every frame now, but the FPS limiter will still be disabled, so it shouldn't take too long if you have a good GPU.

When done, post links to both logs in a comment here. Also post your system specs.

**I need to know**:

* Your OS (and distribution and version, if applicable)
* CPU architecture, and whether you cross-compiled for another one (in that case, which)
* Compiler and its version


Akaricchi:
@laochailan also, I'd like you to test the RNG again, this time from this branch to ensure there were no regressions. And preferably on a larger sample size than what our old test routine spits out.

laochailan:
It’s still working. I also added the script for generating the plots to the repo.

I’ll test the replay later. :)

laochailan:
I’m testing it at the moment. SANIC=-1 went through without a problem.
http://pastebin.com/37WfPp80

SANIC=1 won’t disable the frame limiter for me (probably because of forced v-sync). 
When a dialog appears, it isn’t paged. Instead the character moves around and keeps shooting.
After a long time, the dialog is paged all at once.
http://pastebin.com/raN3zVck

It’s saying OK all the time so I guess this is the unique kind of gameplay recorded in this replay.

```
$ uname -a
Linux youmu 4.9.6-1-ARCH #1 SMP PREEMPT Thu Jan 26 09:22:26 CET 2017 x86_64 GNU/Linux
$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-pc-linux-gnu/6.3.1/lto-wrapper
Target: x86_64-pc-linux-gnu
Configured with: /build/gcc/src/gcc/configure --prefix=/usr --libdir=/usr/lib --libexecdir=/usr/lib --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=https://bugs.archlinux.org/ --enable-languages=c,c++,ada,fortran,go,lto,objc,obj-c++ --enable-shared --enable-threads=posix --enable-libmpx --with-system-zlib --with-isl --enable-__cxa_atexit --disable-libunwind-exceptions --enable-clocale=gnu --disable-libstdcxx-pch --disable-libssp --enable-gnu-unique-object --enable-linker-build-id --enable-lto --enable-plugin --enable-install-libiberty --with-linker-hash-style=gnu --enable-gnu-indirect-function --disable-multilib --disable-werror --enable-checking=release
Thread model: posix
gcc version 6.3.1 20170109 (GCC) 
```

By the way, could it be that Lunatic is not beatable? Like stage 3 and stage 6 and parts of stage 5? Maybe more than parts of stage 5 and a significant amount of stage 4 too?
Rumor has it that stage 3 isn’t even possible on easy. :stuck_out_tongue:

Akaricchi:
Thanks for testing it.

>When a dialog appears, it isn’t paged. Instead the character moves around and keeps shooting. After a long time, the dialog is paged all at once.

I'll investigate the dialogs further (it's possible that I broke something there), but this is probably because when I recorded without a FPS limiter, I waited a few seconds before paging the dialog with the skip key -- which translates into a long wait when you play it back at 60 FPS.

If it says OK all the time, then it's infinitely probable that it, in fact, never desynced -- which is great. However, your system is basically identical to mine, so there's still a chance it'll desync on a different configuration.

I'm not sure if there's a way to reliably disable vsync. If your video driver forces it, there's probably not much to do. That said, we shouldn't even need to test with SANIC set to anything other than -1 if the drawing code never affects the game state. In the master branch it does in some places, which I've fixed, but I wasn't entirely sure that I got all of them down.

>By the way, could it be that Lunatic is not beatable? Like stage 3 and stage 6 and parts of stage 5? Maybe more than parts of stage 5 and a significant amount of stage 4 too?
>Rumor has it that stage 3 isn’t even possible on easy. :stuck_out_tongue:

I'd say the game overall needs some very serious rebalancing in almost every aspect, but that's for another day. I will definitely be tweaking stage 3 first. I remember beating every individual part of stage 3 when I was working on it (yes, even the lunatic version of wriggle night ignite), but probably never as a whole.

I don't think we got the difficulty setting right either. In some places the difference is almost unnoticable. It should be a lot more distinct, e.g. Easy has to be managable by someone who has never played this kind of game before and Hard has to provide some challenge even for an experienced player.

makise-homura:
Tested it!
TAISEI_SANIC=1 looks like totally desynced, as laochailan said. Dialogs paged very accidentally (maybe on 'shoot' button released and pressed again?), player shoots while dialog is on-screen; sometimes player acts totally random (shoots empty places, tries to run through edge of screen, shoots while survival spell is active, etc). Strange that replay has been played to the end: similar issue with demo lump desync, for example, in DOOM causes a demo to end prematurely. Maybe here we have the opposite: replay is somehow 'longer' than game thinks it should be, not 'shorter'.

Two logs differ only in line 84 ("frameskip = ...").
TAISEI_SANIC=-1: http://pastebin.com/72Dyf9eE
TAISEI_SANIC=1: http://pastebin.com/tSDnUzdv

CPU: I think it will not make anything clear, but it is a very non-widely used thing called 'Elbrus-8S' (which have 'Elbrus-2000 v4' 64-bit LE architecture, some kind of VLIW like Intel Itanium). Taisei was built by native compiler.
OS: Something called 'Elbrus OS' which is Debian 8 built for Elbrus architecture.
Compiler: LCC (e[L]brus [C] [C]ompiler) based on gcc 4.8.

`uname -a`:
`Linux sakuya 3.14.79-13.14-e8c #1 SMP Sun Nov 20 04:16:07 UTC 2016 e2k E8C PC-401 GNU/Linux`

`gcc -v`:
`lcc:1.21.12:Nov-17-2016:e2k-v4-linux
Thread model: posix
gcc version 4.8.0 compatible.`

> By the way, could it be that Lunatic is not beatable? Like stage 3 and stage 6 and parts of stage 5?

For me, it looks like non-beatable (without IDDQD) even on easy, lol. =)
Compared to Imperishable Night or Embodiment of Scarlet Devil, Taisei is much more _lunatic_ even on easy modo. =)

laochailan:
This setup :laughing: you must have a lot of fun in your life.

Yes it looks really broken, but those checksums check the value of tsrand and the score. That should be pretty bullet proof. You have to take into account that Akaricchi was trying to play the game on super speed so dodging or doing anything sensible might have been impossible.

Akaricchi:
@makise-homura @laochailan I wasn't *playing* the game -- I was literally **mashing random buttons** with no FPS limiter until it was over. I'll probably record some proper gameplay a bit later, but mind you that I haven't played touhou in a while and I seriously suck.

I warned about that in advance:

> It was recorded on hyper speed with god mode, so just in case you notice that my actions there don't make sense, they aren't supposed to.

So while it looked desynced, according to those logs I'm happy to say that it wasn't, even on such an unusual system. Thanks for giving it a try!

laochailan:
Maybe we should put up a challenge now for someone to write an optimizer that can generate a no miss clear replay through lunatic mode.

Akaricchi:
I've just tested it on windows, and it doesn't desync there either. This could mean either something great (the desync problems are resolved) or awful (the desyncs are rare and unpredictable). Still looking for someone to test it on a 32bit system though, if we still care about those.

Akaricchi:
> Maybe we should put up a challenge now for someone to write an optimizer that can generate a no miss clear replay through lunatic mode.

It's probably easier to git gut enough to no miss clear lunatic manually :laughing:

Akaricchi:
In case any of you want to try it with some actual gameplay, here it is:

[Me trying to play Easy](https://www.dropbox.com/sh/1jkiho7tb6v9wt3/AAC9PaqrLOb-TCqGIX8PTqUQa?dl=0&preview=taisei_20170207_04-00-14_EET_youmuB_easy.tsr)  
[Me getting my ass handed to me on Lunatic, and this time it's on a gamepad with free axis mode](https://www.dropbox.com/sh/1jkiho7tb6v9wt3/AAC9PaqrLOb-TCqGIX8PTqUQa?dl=0&preview=taisei_20170207_04-00-14_EET_youmuB_easy.tsr)

For the uninitiated: the mode of input featured in the second replay generates different kind of events. Joystick axis motion is recorded instead of up/down/left/right keypresses. Which is why this replay is much larger in size than the other ones.

src/replay.c
```diff
@@ -103,219 +113,146 @@ void replay_event(Replay *rpy, int type, int key) {
...
+	SDL_WriteLE32(file, stg->seed);
+	SDL_WriteU8(file, stg->diff);
+	SDL_WriteLE32(file, stg->points);
+	SDL_WriteU8(file, stg->plr_char);
+	SDL_WriteU8(file, stg->plr_shot);
+	SDL_WriteLE64(file, stg->plr_pos_x._int);
+	SDL_WriteLE64(file, stg->plr_pos_y._int);
+	SDL_WriteU8(file, stg->plr_focus);
+	SDL_WriteU8(file, stg->plr_fire);
+	SDL_WriteLE64(file, stg->plr_power._int);
```
laochailan: Using unions like that is probably against the standard and not very portable. I don’t know the details of floating point number representation, but there is probably a better solution.

Akaricchi: As far as I understand it, it's not against the standard in C99 and above (but it is in C++). Assumptions are made that IEEE 754 floating point is used, and endianness is the same for floating point and integral types. In practice both will hold true on any modern system that could ever hope to run Taisei, and even if that wasn't the case, I'd bet the replays would desync anyway because of the FP math differences. But if you want to propose a better solution I'll consider it.

Akaricchi: We can probably just dodge this problem. Store power as an int (internally in the game state too, 105 = 1.05 power), round up the coordinates when the stage starts and store them as ints in the replay. Boom, free portability.

makise-homura:
>This setup 😆 you must have a lot of fun in your life.

Well, I have a couple of ARM boxes at home, but they aren't so interesting as Elbrus, lol :)

>I wasn't playing the game -- I was literally mashing random buttons with no FPS limiter until it was over.

If so, then it's OK, I just noticed it as it clearly reminded me of totally similar behavior of desynced demo lumps in DOOM.

>By the way, why do you play as root?

I never hoped that it will remain unnoticed :) The reason is simple - a bit of laziness that prevented me to make everything as it should be and add a non-root user, do everything in its context and use sudo when needed. At least, that's an engineering installation and something bad that root may accidentally do is not a serious problem compared to hardware issues with non-initializable DRAM in some cases, or, at least, some RTC stuff that caused hwclock's year to be 2037 instead of 2017 (really Sakuya's power, lol :).

>Still looking for someone to test it on a 32bit system though, if we still care about those.

I may try to rebuild Taisei (and every companion library) with `-mptr32` (a some kind of `-m32` for Elbrus) and test it on the same setup (if it will make sense).

PR #59, opened 2017-02-15
[WIP] Migrating from OpenAL/ALUT to SDL2_mixer

makise-homura:
For now it's preliminary version of changes, it is not even tested. Looks like I'll test it today's or tomorrow's evening.


Akaricchi:
Hi. I've only tested the sound effects for now. They seem to work fine, but they play at a lower pitch -- most likely because of sample rate mismatch. I wonder if it'd be possible to painlessly resample the files via SDL/mixer APIs? All of our sounds are in 44100hz, though, so if you pass that instead of `MIX_DEFAULT_FREQUENCY` (which is 22050 I believe) to `Mix_OpenAudio`, that should fix it for now.

lachs0r:
I don’t think SDL_mixer does resampling, so we should ensure that our sounds use the proper sample rate.

makise-homura:
Hmm, tried to change frequency to 44100 - got a lot of errors like `ALSA lib pcm.c:7234:(snd_pcm_recover) underrun occurred` (about 3-5 in a second beginning after `Mix_OpenAudio()` call, even before music/sfx loaded) and very glitchy music and sounds. Looks like there is some trouble in support of such sampling rate from SDL_mixer? Or, maybe, it's fault of my sound subsystem (but it looks strange, as OpenAL performed OK).

lachs0r:
Try the obvious thing first. Increase the mixing chunk size used by SDL (last parameter). Many projects seem to be using 2048 or 4096 there. And yes, it’s definitely a problem with your system.

makise-homura:
I've tried changing this (not even in typical range like 256..8192, but even insane values like 8 or 65536) and sample format nearly at first when identifying the cause of such problem, but there were no significant changes.

But! Looks like playing with `/etc/asound.conf` have a great effect on it, and problem was hidden out there. As I encounter such issue either on my Elbrus system and amd64 too, it looks like there were incorrect values in distribution's default config. Well, now I have either a good sound and no such messages in console, and looks like that solved the problem.

I think I'll write this in README.md, as not only me encounter similar problems, as far as I googled for this.

Akaricchi:
>I think I'll write this in README.md, as not only me encounter similar problems, as far as I googled for this.

Alternatively, we could resample our sounds to 22050 if that's what works better on most systems. I imagine there's a reason why that is the default value.

laochailan:
> Alternatively, we could resample our sounds to 22050 if that's what works better on most systems. > I imagine there's a reason why that is the default value.

Actually there is! The SDL_Mixer documentation says it is the default because 44100 uses too much cpu on toasters.

lachs0r:
Please rebase pull request branches instead of merging master. Force pushing to your own feature branch is okay.

makise-homura:
>Alternatively, we could resample our sounds to 22050 if that's what works better on most systems.

That may be a solution also.

>The SDL_Mixer documentation says it is the default because 44100 uses too much cpu on toasters.

Yes, I've seen these word too, but as Elbrus isn't a kind of toaster :) I've tried to identify problem's source.

>Please rebase pull request branches instead of merging master. Force pushing to your own feature branch is okay.

Ok, (I'm just a bit new to git and was not aware of rebase option, just used merge for such things as I did with svn. As for now, I've looked for it and will try to do things as they should be).

makise-homura:
>Alternatively, we could resample our sounds to 22050 if that's what works better on most systems.

That may be a solution also.

>The SDL_Mixer documentation says it is the default because 44100 uses too much cpu on toasters.

Yes, I've seen these word too, but as Elbrus isn't a kind of toaster :) I've tried to identify problem's source.

>Please rebase pull request branches instead of merging master. Force pushing to your own feature branch is okay.

Ok, (I'm just a bit new to git and was not aware of rebase option, just used merge for such things as I did with svn. As for now, I've looked for it and will try to do things as they should be).

makise-homura:
Well, looks like trying to rebase this branch was a total failure of me (commits were copied, not moved from their original locations).
Or, as I figured out later, I have to do rebase before I push updates to remote repository, not after; and maybe that's the cause of such mess.

What will be best thing to do now? Maybe, close this request without merge, create new branch, transfer changes into that branch, and create a new pull request with new branch? Or there's another way to do it in correct way?

lachs0r:
`git reset --hard 3b03808dd63ddd52cf9dcdd6cfde98516db006c5` and try again (`git rebase master`).

lachs0r:
I have rebased this onto master, squashed some commits and also fixed a few issues. It works for me.
Specifically, sound loading did not work because the switch in load_sound_or_bgm always returned NULL, and because it always set both `snd->sound` and `snd->music` (remember they’re union members).

https://github.com/lachs0r/taisei/tree/sdl2_mixer

src/resource/resource.c
```diff
@@ -39,18 +39,26 @@ void recurse_dir(char *path) {
...
+		 || (strcmp(dp->d_name + strlen(dp->d_name)-4, ".mod" ) == 0)
+		 || (strcmp(dp->d_name + strlen(dp->d_name)-4, ".xm"  ) == 0)
+		 || (strcmp(dp->d_name + strlen(dp->d_name)-4, ".s3m" ) == 0)
+		 || (strcmp(dp->d_name + strlen(dp->d_name)-4, ".669" ) == 0)
+		 || (strcmp(dp->d_name + strlen(dp->d_name)-4, ".it"  ) == 0)
+		 || (strcmp(dp->d_name + strlen(dp->d_name)-4, ".med" ) == 0)
+		 || (strcmp(dp->d_name + strlen(dp->d_name)-4, ".mid" ) == 0)
+		 || (strcmp(dp->d_name + strlen(dp->d_name)-4, ".flac") == 0)
+		 || (strcmp(dp->d_name + strlen(dp->d_name)-4, ".aiff") == 0)
+		 || (strcmp(dp->d_name + strlen(dp->d_name)-4, ".voc" ) == 0))
```
Akaricchi: Should at least be using `strendswith()` in here (defined in global.c).

Akaricchi:
@lachs0r tested your branch just now (but just the SFX, again). Seems to work fine, but I get way better results with chunk size set to 2048, and even better with 1024. The delay with larger chunks is especially noticable with the player shooting sound. We have to strike a good balance here. Generally the lowest we can go without causing underruns/skipping on most systems, the better. Maybe expose it as an "advanced" option in the config? Would be nice if we could auto-detect the optimal setting somehow, but I doubt that's even possible.

Speaking of the shooting sound, it is still ear rape with any chunk size, and so was it with OpenAL. I can't stand playing this game with headphones because of it. It really needs a gentler and quiter replacement IMO.

lachs0r:
I guess I’ll just close this in favor of my cleaned-up branch.

@makise-homura if you still want to work on this before it gets merged, you can also send PRs to my branch.

src/resource/resource.c
lachs0r: That’s fixed in my branch. This code also looks like it doesn’t compare 4-character extensions correctly.

makise-homura:
>I have rebased this onto master, squashed some commits and also fixed a few issues. It works for me.

Thank you! It's ok.

>Specifically, sound loading did not work because the switch in load_sound_or_bgm always returned NULL, and because it always set both snd->sound and snd->music (remember they’re union members).

Yes, I've fixed it in my copy (got rid of union and moved `return NULL` to error reaction block), but after this tries to recover after failed rebase these changes were not pushed to my remote branch.

>Should at least be using strendswith() in here (defined in global.c).

Wow, I just wanted some kind of such function for this code. As it really exists, it's the best way to do such comparisons.
(By the way, `resource.c:40` still contains `if(strcmp(dp->d_name + strlen(dp->d_name)-4, ".png") == 0)`, is it left intentionally without fixing to `strendswith()`?)

>This code also looks like it doesn’t compare 4-character extensions correctly.

Yes, that's so (I was planning to test my code at monday with every sound/music format, so this mistake may be found then, but now I see there's no mistake at all).

>Seems to work fine, but I get way better results with chunk size set to 2048, and even better with 1024. The delay with larger chunks is especially noticable with the player shooting sound.

Well, I've noticed a delay with chunk sizes > 8192, but something like 256-4096 was ok. So, I think these parameters require fine-tuning for each variation sound subsystem... But yes, looks like chunk size should be low by default.

>if you still want to work on this before it gets merged, you can also send PRs to my branch.

I planned to improve code of loading sfx/music files to get rid of such copy-pasted lines, but I see you already did it. So well, if I found something to fix or improve in that code while testing it at monday, I'll make a PR to your branch.
Another thing I planned to do is to make sound/sfx support optional (if cmake is not able to find SDL2_mixer at build-time, then it will not fail an entire build, but will turn off sound support, as in GZDoom, for example); but I'm still not sure will it be really useful or not.

lachs0r:
Well, now the change is in master, so no need to send PRs to my branch.

makise-homura:
Ok, I see. Then I'll continue to make PRs to master repository.

PR #61, opened 2017-02-17
Config refactor: callbacks and macro hell

Akaricchi:
I was going to push this straight into master initially, but the changes ended up being a little too big and will probably cause plenty of conflicts with #59. Among other things, I've added a possibility to hook up callbacks on config options to watch for changes without having to hack the options menu code, so any such things that would init/uninit audio, or set volume, etc. have been moved from there.

What do you think, @makise-homura, should I merge this now (then you'll have to rebase and resolve the conflicts by yourself), or wait for your PR to be merged? I prefer the former option of course, but I don't know if you have that kind of time.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -7,181 +7,136 @@
  */
 
 #include <string.h>
+#include <assert.h>
 
 #include "config.h"
 #include "global.h"
 #include "paths/native.h"
 #include "taisei_err.h"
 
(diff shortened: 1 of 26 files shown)


makise-homura:
As far as I see, there are no serious conflicts except config values read/write, so I think, I'll resolve them by the way.

Issue #66, opened 2017-02-24
Don’t load everything in the beginning.


laochailan:
We are having a pretty long loading time now. It will only get worse with BGM in place. So what about somehow switching to loading things stage-wise.

At the moment most of the time is spent loading the big spell backgrounds and textures for the later stages.

Those resources are already in special folders like `gfx/stage6/`, so maybe we could add something that dynamically loads and unloads the right folder them to the stage logic.
This would also make testing spells in single spell mode faster.

Akaricchi:
Such a system should be flexible enough to preload resources for all the stages when the story mode starts (and then release them), and definitely no reloading every time one restarts a stage (think spell practice). Lag at stage transitions should be minimized.

laochailan:
In the original Touhou you also have a little loading thing at stage transitions and it doesn’t hurt because you need some time to relax between transitions too.

Maybe the loader shouldn’t release them or release them very hesitantly so you don’t have to reload every time you enter a stage. Finally we could make it load asynchronously and add like hints when to start loading stuff (e.g. immediately after the boss explodes).

It may be a bit complicated, but the way it is now you have to wait 3 seconds every time just to load the stage6 assets which you really never need. They could also be compressed in a lossy way.

Akaricchi:
>In the original Touhou you also have a little loading thing at stage transitions and it doesn’t hurt because you need some time to relax between transitions too.

We do give time already, our stages don't end immediately after the boss and don't start with a bullet explosion. I think it's good to have the "relax time" be consistent this way rather than making it vary with loading times. Just because Touhou has a mid-stage loading screen doesn't mean we should. I believe the best kind of loading is one you don't notice.

Asynchronous loading is certainly an attractive option, but it does impose some challenges. Pretty much none of our code is thread-safe right now so extra care will have to be taken when implementing it. But I'm not opposed to the idea if we can get it right.

This is my intermediate proposition. I think this should clean up the stage and resources related code a bit and make it easier to implement asynchronous loading later.

**Phase 1: stage mini-refactoring**

- **Get rid of the `stage*_loop` functions.** Define a `StageProcs` struct that would hold all the stage-specific function pointers that we currently pass to `stage_loop`.
- **Have `StageInfo` contain a pointer to `StageProcs`** rather than a pointer to a loop function.
- **Move handling of stage BGM and end timers to the stage event handlers**. E.g. `AT(0) play_bgm`; `AT(5000) end_stage`.
[...]

PR #67, opened 2017-02-28
Implemented transient and permanent resources (issue #66)

makise-homura:
Looks like I did the stuff according to the first approach of "Phase 2" of @Akaricchi in issue #66.

What have been done:
- introduced `ListEx` (`List` with `transient` attribute, should be third variable in list item) and `delete_all_or_transient_elements()` that looks on it and does what its name means;
- resource name determination (`determine_name_with_path()` for textures and `determine_name()` for everything else);
- stage now has `resource_id` that specifies which stage transient resources pack should be loaded (determined by `global.stage->resource_id`). It is equal to `id` in "base" stages, and equal to its ancestor's `id` in descendant stages (such as spell practice ones);
- got rid of "resource type loaded" states, loading now based on real existence of so-named resource in list (if there is one, new resource with this name is not loaded), `get_..._quiet()` functions thereby added that do not say "resource not found" as normal `get_...()` do;
- `recurse_dir()` now uses relative path and decides what directory does contain: transient or permanent resources. For now it is done by relative path (`strstr(path, "stage")`). It traverses every suitable directory, but loads only resources of specified type;
- some minor standartization of shutting down sfx and bgm subsystem;
- minor fixes.

As a result, startup time decreased from 7-10 seconds to 3-4. Looks like good thing. Interstage delays are still not noticeable, then.

What to do:
- Test it, of course (I just tried to view 6-stage replay, turn on/off music/sound/shaders, tried some stage practice) and encountered no problems.
- Sort resources related to stages by corresponding subdirectories (it may be done freely for everything except textures; texture names should be also fixed in source code, as their names are based not only on their filenames, but on relative path too).
- Think what to do with "always permanent" resources that even may not be based on files (VBOs, FBOs, fonts, BGM descriptions, shader snippets) and their "load"/"unload" (initialization/deinitialization). Maybe nothing.
- Think if we need transiency as in @Akaricchi's method (by a list), or @laochailan's one will be enough.
- Implement transient -> permanent switchover when resource of same name is loaded as permanent.
- Implement async loading (if required).

P.S. Stage 4 used one of the resources (`border.png`) of Stage 2; for now, I just copied it from stage 2 subdirectory to stage 4, but may there be any another ways to do it, such as make this resource permanent, not transient.


Akaricchi:
I've started working on a branch that changes the resources code to use hashtables instead of linked lists before I saw this pull request. I have pushed it just now and going to open my own pull request soon. Maybe you could implement it on top of that instead? I don't think linked lists is a good way to implement a name->value mapping, and hacking another attribute into them doesn't seem like a good idea either.

Also, maybe we should think about generalizing our resource handling in some way. E.g. have a structure that looks about like this:

```
typedef enum ResourceType {
    RES_TEXTURE,
    RES_SOUND,
    ......
};

typedef struct Resource {
    ResourceType type;
    bool transient;

    union {
        Texture *texture;
        Sound *sound;
        ......
    };
};
```

And then functions like `Resource* get_resource(ResourceType type, char *name, bool transient)`. This way the loading code for each individual resource type won't have to care about whether we want it transient or not, or whether the resource is already loaded, or about putting it in a hashtable, etc.

Also, why would you replace `#include <SDL_opengl_glext.h>` with `#include <GL/glext.h>`? This is not portable and there shouldn't be a reason to do that.

Akaricchi:
Here's the hashtables branch/PR #68

Akaricchi:
Also, please join us on IRC (#taisei-project @ irc.freenode.net). It's easier to collaborate with real-time communication.

laochailan:
Yes, I’m also not a fan of introducing a `ListEx` and the other things Akaricchi has mentioned.

But thanks for your work! Maybe we should have coordinated this better. We are having a huge part of the discussion on irc #taisei-project on freenode. If you want to join the dev talk and know in detail what we are working on, just stop by.

makise-homura:
> Also, please join us on IRC (#taisei-project @ irc.freenode.net).
Well, I'll try now to join.

> Also, why would you replace #include <SDL_opengl_glext.h> with #include <GL/glext.h>?
Oops, that was not related to this issue. It's because my SDL2 headers somehow do not include this file (either on Elbrus and x86_64 machines). I have an idea to investigate it later (as I see some python-based magic in this file that I was not trying to understand for now), but now it's just quick fix to build Taisei on my setup.

> Maybe you could implement it on top of that instead?
Maybe. But how do we do this? Should I rebase my branch to "hashtable" and then review my changes then? (And is it possible to change destination branch of PR?) Or should I wait when PR #68 be in master and then do the same? (Hope we will talk about it on IRC then).

> I don't think linked lists is a good way to implement a name->value mapping, and hacking another attribute into them doesn't seem like a good idea either.
As far as I know, hashtable usage give a bit more universality to "name->value" mapping instead of list, and I thought of using it, but considered it a bit overkill for us (they offer a great help when item count grows up to thousands and more). But if they are implemented, that's right way to use them, of course.

> Also, maybe we should think about generalizing our resource handling in some way.
Yes, that's good idea. For example, having a bunch of functions like get_something(...) with mostly similar code embarrassed me a bit.

Akaricchi:
>Oops, that was not related to this issue. It's because my SDL2 headers somehow do not include this file (either on Elbrus and x86_64 machines). I have an idea to investigate it later (as I see some python-based magic in this file that I was not trying to understand for now), but now it's just quick fix to build Taisei on my setup.

That probably means your SDL2 version is just old (which?) and you should consider updating. The latest stable release definitely has it. https://github.com/spurious/SDL-mirror/blob/master/include/SDL_opengl_glext.h

>Maybe. But how do we do this? Should I rebase my branch to "hashtable" and then review my changes then? (And is it possible to change destination branch of PR?) Or should I wait when PR #68 be in master and then do the same? (Hope we will talk about it on IRC then).

You could rebase, but you will obviously get a lot of merge conflicts. I'm not sure what'd be easier for you, resolve them or just reimplement the functionality from scratch based on that branch. Do whatever is more comfortable for you.

Yes, you can change the destination branch, press the `Edit` button right to the title. But maybe in this case it'd be cleaner to just open a new pull request.

makise-homura:
So, closing this, because this should be fully redone after @Akaricchi finish his own resource management subsystem.

PR #68, opened 2017-02-28
Hashtables and resources semi-refactoring

Akaricchi:
This adds a little general purpose hashtable library and replaces the linked lists in resources code with hashtables.

### DIFF
--- a/src/resource/audio.c
+++ b/src/resource/audio.c
@@ -201,8 +201,6 @@ void set_sfx_volume(float gain)
 	Mix_Volume(-1, gain * MIX_MAX_VOLUME);
 }
 
-#include <assert.h>
-
 void* delete_sound(void *name, void *snd, void *arg) {
 	Sound *ssnd = (Sound *)snd;
 
--- a/src/resource/audio.c
+++ b/src/resource/audio.c
@@ -114,7 +114,7 @@ void shutdown_sfx(void)
 	unload_mixer_if_needed();
 }
 
-Sound *load_sound_or_bgm(char *filename, Sound **dest, sound_type_t type) {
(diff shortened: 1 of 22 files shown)


src/resource/audio.c
```diff
@@ -205,14 +201,21 @@ void set_sfx_volume(float gain)
 	Mix_Volume(-1, gain * MIX_MAX_VOLUME);
 }
 
-void delete_sound(void **snds, void *snd) {
+#include <assert.h>
```
laochailan: Should this be here?

src/resource/model.c
```diff
@@ -215,17 +213,16 @@ void draw_model_p(Model *model) {
...
 	free(((Model *)model)->indices);
-
-	delete_element(models, model);
+	free(model);
+	return NULL;
 }
 
-void delete_models(void) { // Does not delete elements from the VBO, so doing this at runtime is leaking VBO space
-	delete_all_elements((void **)&resources.models, delete_model);
+void delete_models(void) { // Does not delete elements from the VBO, so doing this at runtime is leaking VBO spac
```
laochailan: That *e* got missing.

src/resource/shader.c
```diff
@@ -309,33 +308,24 @@ int uniloc(Shader *sha, char *name) {
...
-		if(strcmp(s->name, name) == 0)
-			res = s;
-	}
-
-	return res;
+	return hashtable_get_string(resources.shaders, name);
 }
 
-void delete_shader(void **shas, void *sha) {
-	free(((Shader *)sha)->name);
```
laochailan: I don’t see where the shader name is freed in the new version.

PR #70, opened 2017-03-06
Major refactoring

Akaricchi:
This is mostly about the resources and audio subsystems, which have been largerly redesigned and rewritten. It also adds some utility functions (and moves them into a separate module), enables OpenGL debugging in debug builds, and removes some legacy cruft such as NO_SHADER/NO_AUDIO, among other things.

I have not yet tested this on OSX or Windows, except for just cross-compiling the executables for them.

@makise-homura, you can probably start working on deferred loading based on this branch, if you still want to. I may update it a few more times before it's merged, but there shouldn't be any too-drastic changes.


src/util.c
```diff
@@ -117,6 +117,18 @@ char* copy_segment(const char *text, const char *delim, int *size) {
...
+
+    while(true) {
+        c = strchr(buf, '\0') - 1;
+        if(*c == '/' || *c == '\\')
+            *c = '\0';
+        else
+            break;
+    }
+}
+
```
laochailan: what about
``` c
for(char *c = buf[strlen(buf)-1]; c >= buf && *c == '/'; c--)
    *c = 0;
```

PR #72, opened 2017-03-09
Deferred loading of resources

makise-homura:
First work on first part of deferred (and then async) resource loading.
- load_resources() now skips transient resources;
- free_resources() now gets an argument which may include RESF_TRANSIENT, so it frees only transient resources;
- a callback specific for each resource type used to determine if resource transient or permanent (NULL as callback means resources of that type are always permanent);
- resources now manually loaded by a hardcoded calls to get_resource() at preload() callback on stage start;
- transient resources freed at stage end.

TODOs:
- make some models, animations, dialog gfx transient;
- introduce this functionality to spell stages;
- maybe load menu, credits and ending stuff as transient also.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -26,7 +26,7 @@ void taisei_shutdown(void) {
 	printf("\nshutdown:\n");
 
 	free_all_refs();
-	free_resources();
+	free_resources(0);
 	audio_shutdown();
 	video_shutdown();
 	gamepad_shutdown();
--- a/src/resource/resource.c
+++ b/src/resource/resource.c
@@ -37,6 +37,7 @@ static void register_handler(
 	ResourceNameFunc name,
 	ResourceFindFunc find,
 	ResourceCheckFunc check,
+	ResourceTransientFunc istransient,
 	size_t tablesize)
 {
 	assert(type >= 0 && type < RES_NUMTYPES);
@@ -48,6 +49,7 @@ static void register_handler(
 	h->name = name;
 	h->find = find;
 	h->check = check;
+	h->istransient = istransient;
 	h->mapping = hashtable_new_stringkeys(tablesize);
 	strcpy(h->subdir, subdir);
 }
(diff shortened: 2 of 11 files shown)


Akaricchi:
Looks ok overall. But we will have to remove the `istransient` callbacks eventually, because we've decided to not rely on directory scanning at all and load all of our resources explicitly instead. So there won't be any use for those functions when that's implemented. But I guess it's ok for an intermediate solution.

For the spell stages, you can just use the same preload function as for the main stage in `stageX_spell_procs`.

The call to `stage->preload` should probably come before the call to `stage_start`. Maybe right after the assignment to `global.stage`.

Transient resources shouldn't be freed when the stage ends. The caller of `stage_loop` should be expected to clean them up, e.g. the `game_start` and `replay_play` functions. This approach is more flexible because it allows us to, for example, not reload resources every time the player restarts a stage (which can be very often if it's a spell stage), or preload all of the 1-6 stages when starting up the story mode (to eliminate transition delays until async loading is implemented).

Akaricchi:
Also, maybe transient should be the default mode and the flag should be `RESF_PERMANENT`. So that no one can accidentally load a resource that should be transient in permanent mode (in stage code or the `preload` functions). Right now, wrapper functions like `get_tex` all call `get_resource` without `RESF_TRANSIENT`, which means they'll load the resource in permanent mode if it was not explicitly preloaded before.

Probably the same thing should be done about `RESF_REQUIRED`: make it the default and add a `RESF_OPTIONAL` instead. That would make more sense, because most of our resources are not optional. But that was my own oversight, so you can leave fixing that up to me if you want.

Akaricchi:
I have started working on removal of resource scanning and further developing this in the `resources` branch. Your changes have been integrated there.

Akaricchi:
Closing this because I've merged it in #73 and developed it further there. You can send pull requests into that branch if you want.

PR #75, opened 2017-03-15
Asynchronous resource preloading

Akaricchi:


src/hashtable.c
```diff
@@ -52,12 +59,39 @@ Hashtable* hashtable_new(size_t size, HTCmpFunc cmp_func, HTHashFunc hash_func,
...
+
+static void hashtable_enter_state(Hashtable *ht, int state, bool mutex) {
+    while(!hashtable_try_enter_state(ht, state, mutex));
+}
+
+static void hashtable_idle_state(Hashtable *ht) {
+    SDL_AtomicSet(&ht->cur_operation, HT_OP_NONE);
+}
+
+void hashtable_lock(Hashtable *ht) {
```
laochailan: The code in this file doesn’t use the lock/unlock aliases for the most part. I think enter_state/idle_state should be renamed to lock/unlock maybe to make clearer what they do and that for every enter_state you need an idle_state.

src/resource/resource.c
```diff
@@ -94,9 +100,99 @@ static char* get_name(ResourceHandler *handler, const char *path) {
...
+
+static bool resource_check_async_load(ResourceHandler *handler, const char *name) {
+	if(SDL_ThreadID() == main_thread_id)
+		update_async_load_state();
+	ResourceAsyncLoadData *data = hashtable_get_string(handler->async_load_data, name);
+	return data;
+}
+
+static void resource_wait_for_async_load(ResourceHandler *handler, const char *name) {
+	// XXX: is there a better way than a busy loop?
```
laochailan: An async loop with SDL_Delay?

src/hashtable.c
Akaricchi: It doesn't use them here at all. Notice how they are not static. They are meant to be used by external code with the otherwise thread-unsafe iterators, which may not be the best interface, but at least it abstracts away the whole internal state business.

src/resource/resource.c
Akaricchi: Wouldn't that just make it less responsive? That comment is kinda old/irrelevant, I actually think a busy loop is ok for such short-ish waits now, but correct me if I'm wrong.

laochailan: Yes sure, but if the waits are short, I guess it’s okay.

src/hashtable.c
laochailan: I guessed that, maybe it’s better that way. I’m not a specialist on this matter. I also don’t understand what the mutex argument does.

Akaricchi: It stands for "mutual exclusion". If `mutex` is true, the call to `hashtable_enter_state` will block until the state is `HT_OP_NONE` (effectively implementing a [critical section](https://en.wikipedia.org/wiki/Critical_section)). If it's false, it will also do that, but only if the current state is different from the state we want to enter (and is not `HT_OP_NONE`, of course).

`mutex` is true for operations that modify the hashtable, and false for those that only read from it. This allows multiple threads to read from the hashtable at the same time, but forbids concurrent writes, reads during writes, and writes during reads. Since reading from a hashtable ought to be more common than writing to it, I think it's good to get rid of unnecessary blocking there.

I previously used [SDL's mutexes](https://wiki.libsdl.org/SDL_CreateMutex) there for synchronization, but they aren't flexible enough to enable the exact behaviour I wanted (concurrent reads), so I rolled my own version using the atomic primitives.

Akaricchi: hmm, now that I think of it again, there might be a problem with concurrent reads. If one thread sets state to `HT_OP_NONE` while others are still reading, then there's a tiny time window where a write can still occur during a read, which is not good. I'll look into fixing that soon. Probably should use some sort of counter/semaphore there.

PR #79, opened 2017-04-01
C++ rewrite

Akaricchi:
Because it's obviously the superior language, duh.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -112,6 +112,8 @@ else()
 	)
 endif()
 
+set_source_files_properties(${SRCs} PROPERTIES LANGUAGE CXX)
+
 if(DEFINED TAISEI_DEBUG)
 	if(TAISEI_DEBUG)
 		set(CMAKE_BUILD_TYPE "Debug")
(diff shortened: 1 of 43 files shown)


lachs0r:
I think we should use some of [Boost’s libraries](http://www.boost.org/doc/libs/) and some STL features like `shared_ptr`, `hash_map`, etc.

And switch to [SFML](https://www.sfml-dev.org/), which takes a more object-oriented approach than SDL (which is really showing its age).

laochailan:
If we do a rewrite we should go it properly and switch to Unity altogether. No need to reinvent the wheel.

It supports all the platforms we care about and we could use the integrated physics support for realistic Projectile motion.

Akaricchi:
Unity is of course a great engine, but it has an unfortunate reputation among avid gamers, which could paint us in bad light. We obviously want to look very serious and professional, and for that goal, I believe the right choice to be the latest and greatest Unreal Engine 4. We also should consider integrating Scaleform to build a rich and innovative user interface, and Simplygon to optimize the 3D performance.

While we're at it, I think we should stop supporting platforms with very little market share like Lunix. They clearly represent a very insignificant portion of our userbase. Dropping the support would greatly reduce the development costs and allow us to focus on the more relevant and widespread platforms to expand our target audience, such as Apple's iOS.

Last but not least, we really have to keep up with the times. VR is obviously the present and future of gaming, and we can not ignore it. Anyone wants to buy me an oculus rift? My birthday is in 8 days from now, thank you.

laochailan:
Agreed, our Linux user base is like at most 3 users.

Concerning the Oculus Rift, what about starting a Kickstarter campaign?

Akaricchi:
april fooooolsxxxXxDDDDdddddddddlolololol:^)

PR #82, opened 2017-05-06
Virtual filesystem

Akaricchi:
This is an abstraction layer over filesystem access. Portability, virtual directories, unions, transparent zip archive support, blahblahblah the cool stuff.

Seems mostly ready but needs more testing and probably documentation. Code review is appreciated too.

This introduces a new dependency, libzip. Fortunately it's a widely available and popular library, so it shouldn't be a pain in the ass to depend on. I am thinking of switching to libarchive instead, though, since it supports a lot more formats and seems to be just as popular.


src/vfs/syspath_win32.c
```diff
@@ -0,0 +1,280 @@
...
+    .free = vfs_syspath_free,
+    .locate = vfs_syspath_locate,
+    .syspath = vfs_syspath_syspath,
+    .iter = vfs_syspath_iter,
+    .iter_stop = vfs_syspath_iter_stop,
+    .mkdir = vfs_syspath_mkdir,
+    .open = vfs_syspath_open,
+};
+
+static void vfs_syspath_normalize(char *path) {
```
laochailan: Is this necessary? I mean who cares about a double separator in these and forward slashes are valid on windows right?

src/main.c
```diff
@@ -101,6 +100,112 @@ static void init_sdl(void) {
...
+	}
+
+	if((*storage = getenv("TAISEI_STORAGE_PATH")) && **storage) {
+		*storage = strdup(*storage);
+	} else {
+		*storage = SDL_GetPrefPath("", "taisei");
+	}
+}
+
+static bool filter_zip_ext(const char *str) {
```
Akaricchi: Somehow I haven't even thought about other extension checking. It's not critical since we only use it for our resources so far (all of which are lowercase), but you're right that'd be more correct.

Maybe we should add a case-sensitivity parameter to the str{starts,ends}* functions instead, to not duplicate code. strendswith_any is also used for extension testing.

src/main.c
```diff
@@ -101,6 +100,112 @@ static void init_sdl(void) {
...
+	struct mpoint_t {
+		const char *dest;	const char *syspath;							bool mkdir;
+	} mpts[] = {
+		{"storage",			storage_path,									true},
+		{"resdirs",			res_path,										false},
+		{"resdirs",			p = strfmt("%s/resources", storage_path),		true},
+		{NULL}
+	};
+
+	vfs_init();
```
Akaricchi: @laochailan I think it's a naming issue. vfs_init is part of the VFS API, it sets up the root and  other bare minimum things required to use the VFS in any way. init_vfs is a consumer function, it calls vfs_init and configures the VFS for use in Taisei. So maybe something like "setup_vfs" would be a more appropriate name for it?

src/main.c
```diff
@@ -101,6 +100,112 @@ static void init_sdl(void) {
...
+		free(tmp);
+	}
+
+	vfs_dir_list_free(ziplist, numzips);
+
+	vfs_mount_alias("res", "respkgs");
+	vfs_mount_alias("res", "resdirs");
+
+	vfs_unmount("resdirs");
+	vfs_unmount("respkgs");
```
Akaricchi: The idea here is to mount ZIPs from all the "real" resource directories (in our case, the system-wide one (or ./data/ with RELATIVE) and the user's override in $XDG_DATA_HOME/taisei/), but still give precedence to any plain local file over any file within a zip.

Yes, I realize it's not an important detail for us - at least for now without custom spell scripts :^)

Tell me if you want me to elaborate how this thing works. Or maybe I should leave some helpful comments there.

src/util.c
```diff
@@ -40,6 +40,16 @@ bool strendswith_any(const char *s, const char **earray) {
     return false;
 }
 
+bool strstartswith_any(const char *s, const char **earray) {
```
Akaricchi: Good catch, but it's not misnamed, it's broken... strendswith_any just above it is exactly the same. Ctrl+C/Ctrl+V programming :/

It's not used anywhere yet, I just added it for parity. I probably shouldn't do that without testing the thing at least.

src/util.c
```diff
@@ -129,6 +139,16 @@ void strip_trailing_slashes(char *buf) {
         *c = 0;
 }
 
+char* strappend(char **dst, char *src) {
```
Akaricchi: I don't know, such misuses should at least raise some eyebrows, because:

- This function takes a `char**`, meaning it can change the value of the pointer. No function that can operate on stack-allocated arrays may do this.
- `strcat` and `strncat` exist, and they of course take a `char*`.

I don't know if it can be made any safer, other than explicitly documenting it -- which'd be totally fine, but outside the scope of this PR since lack of documentation is an ubiquitous problem in this codebase.

src/util.h
```diff
@@ -34,15 +35,17 @@
...
 char* copy_segment(const char *text, const char *delim, int *size);
-bool strendswith(const char *s, const char *e)  __attribute__((pure));
-bool strstartswith(const char *s, const char *p)  __attribute__((pure));
-bool strendswith_any(const char *s, const char **earray)  __attribute__((pure));
+bool strendswith(const char *s, const char *e) __attribute__((pure));
+bool strstartswith(const char *s, const char *p) __attribute__((pure));
+bool strendswith_any(const char *s, const char **earray) __attribute__((pure));
+bool strstartswith_any(const char *s, const char **earray) __attribute__((pure));
 void stralloc(char **dest, const char *src);
 char* strjoin(const char *first, ...) __attribute__((sentinel));
```
Akaricchi: As I said above, I'm all for it but that's a separate issue. Current rule of thumb is, if it returns a `char*` and not a `const char*`, it probably should be free'd. There are some exceptions to this, though (e.g. vfs/pathutils.c).

src/vfs/syspath_win32.c
Akaricchi: It absolutely isn't, I just hate ugly inconsistent paths in logs. If you're worried about potential bugs in this function, perhaps this should be done at the representation level (i.e. in `vfs_syspath_syspath`)?

Akaricchi:
There have been no apparent issues on any platform for a while, so I'm merging this.

Issue #83, opened 2017-05-07
The UBSan detection is unreliable


Akaricchi:
It gives a false positive on two of my Fedora systems: 25 stable on my netbook, and 27 rawhide in a systemd-nspawn container. 

[The build log](https://gist.github.com/Akaricchi/728c0c774595bb6bd1be20f77fa1c838)

Akaricchi:
Pinging @makise-homura

makise-homura:
Wow.
Is that possible to see CMakeFiles/CMakeOutput.log and CMakeFiles/CMakeError.log, if latter exists?

Akaricchi:
Here's the relevant portion of CMakeOutput.log:

```
Performing C SOURCE FILE Test COMPILER_SUPPORTS_UBSAN succeeded with the following output:
Change Dir: /home/akari/taisei/build/CMakeFiles/CMakeTmp

Run Build Command:"/usr/bin/gmake" "cmTC_615b3/fast"
/usr/bin/gmake -f CMakeFiles/cmTC_615b3.dir/build.make CMakeFiles/cmTC_615b3.dir/build
gmake[1]: Entering directory '/home/akari/taisei/build/CMakeFiles/CMakeTmp'
Building C object CMakeFiles/cmTC_615b3.dir/src.c.o
/usr/bin/cc   -DCOMPILER_SUPPORTS_UBSAN   -fsanitize=undefined -o CMakeFiles/cmTC_615b3.dir/src.c.o   -c /home/akari/taisei/build/CMakeFiles/CMakeTmp/src.c
Linking C executable cmTC_615b3
/usr/bin/cmake -E cmake_link_script CMakeFiles/cmTC_615b3.dir/link.txt --verbose=1
/usr/bin/cc  -DCOMPILER_SUPPORTS_UBSAN    -rdynamic CMakeFiles/cmTC_615b3.dir/src.c.o  -o cmTC_615b3 
gmake[1]: Leaving directory '/home/akari/taisei/build/CMakeFiles/CMakeTmp'

Source file was:
int main(void) { return 0; }
```

You can see here that it does not pass `-fsanitize=undefined` to the linker command, which is required for UBSan to work. Fedora (and possibly other distros) does not ship libubsan and libasan with GCC by default, they are separately installable packages. Thus the compilation succeeds, because the compiler supports the flag, but the linking fails.

makise-homura:
[...]

PR #84, opened 2017-05-10
Some compartibility propositions

makise-homura:
- Fixing issue #83 (another attempt, added specific cmake script to make parameter checks the way we really want, not the way that legacy cmake modules do)
- Reintroducing ASAN check in addition to UBSAN, as we may do it now
- Added checking for -Wpedantic and -Wnull-dereference for older compilers


Akaricchi:
Actually curious, what relevant compiler doesn't support `-Wpedantic`? Some really old GCC versions? If that's a problem, we could just revert it to `-pedantic`. That's an older (now deprecated) flag that does the same thing.

I'm not sure about the `-Wnull-dereference` check either. It's not a very important warning so I'd rather just remove it. But an ideal solution would be to somehow make GCC just ignore unrecognized flags. Clang just prints a message, for example. This way we could enable some useful GCC7 and clang warnings too (switch fallthrough and local shadowing, for example).

makise-homura:
>That's an older (now deprecated) flag that does the same thing.
Maybe use `-Wpedantic` if check succeeds, and `-pedantic` otherwise? To not broke compartibility with those compilers that do not support `-Wpedantic`, but use non-deprecated thing if compiler supports it.

>But an ideal solution would be to somehow make GCC just ignore unrecognized flags.
Well, that would be ideal, but I think that's some king of gcc philosophy like "reject everything non-understandable and forbid changing this behavior". I see a lot of places where it will be a lot easier to just do something like `CFLAGS=-fignore-unknown-params make all` instead of grepping makefiles for unknown options, but unfortunately it looks like gcc is strictly against this way...

laochailan:
Do we need that level of compatibility for debug builds? If the release builds work on every toaster that is okay, but maybe we can expect developers to have a modern compiler (or clang). Our cmake file is already very complicated.

Alternatively create a SuperUberDebug build type and move the new stuff there.

makise-homura:
>Our cmake file is already very complicated.

I have seen build configuration system for Intel DPDK once, so after that I don't see anything else as too complicated. :)

>Alternatively create a SuperUberDebug build type

That's another way. Because debug (when you try to figure out what's going on on your system when Taisei runs) and source checking (when you enable all warnings and stuff) are a bit different modes.

makise-homura:
I just moved checking stuff to a separate module and made some cleaning in src/CMakeLists.txt. Now these checks looks there quite more simple than before.

If there'll be need in another compile/link-time parameter checks, this module may be used for this too, I hope.

laochailan:
The way you did looks neat. :)

Kind of sad one has to do all this boilerplate stuff just to do something that should be somehow cmake’s job.

makise-homura:
Thanks :) I tried to make this such a way that reusable cmake code should be not in project file, but in a separate module that may be used wherever it will be need for it - as long it is independent from the project specifics.
Also, I wonder too why did not anyone implemented this stuff before in standard cmake modules...

PR #86, opened 2017-09-29
Yet another compatibility propositions

makise-homura:
- Lowered version requirements of CMake;
- Made libzip and usage of -flto flags optional and autodetectable at build time;
- Added DISABLE_ZIP option.


src/CMakeLists.txt
```diff
@@ -10,6 +10,11 @@ endif()
 
 find_package(PkgConfig REQUIRED)
 
+if(CMAKE_VERSION VERSION_LESS 3.3)
+	message(STATUS "  Version of cmake is too old, data packaging will not be used.")
+	set(DISABLE_ZIP ON)
```
Akaricchi: I don't think this should disable ZIP support completely. If anything, it should disable only `PACKAGE_DATA`. But even then, I would prefer that in case `PACKAGE_DATA` is enabled and Cmake is too old, it would exit with an error instead of automatically disabling `PACKAGE_DATA`. The error then would explain that either Cmake should be upgraded to 3.3+, or `PACKAGE_DATA` should be disabled manually. This is consistent with how `PACKAGE_DATA_LEANIFY` is handled.

src/CMakeLists.txt
```diff
@@ -21,12 +26,22 @@ pkg_check_modules(SDL2 sdl2>=2.0.5 REQUIRED)
 pkg_check_modules(SDL2_TTF SDL2_ttf REQUIRED)
 pkg_check_modules(ZLIB zlib REQUIRED)
 pkg_check_modules(PNG libpng>=1.5 REQUIRED)
-pkg_check_modules(ZIP libzip>=1.0 REQUIRED)
+pkg_check_modules(ZIP libzip>=1.0)
+
+if(NOT ZIP_FOUND)
+	message(STATUS "  No suitable libzip found, data packaging will not be used.")
+	set(DISABLE_ZIP ON)
```
Akaricchi: Same thing here. If `libzip` is not found or too old, ask the user to install/upgrade it and exit with an error.

src/CMakeLists.txt
```diff
@@ -163,8 +183,8 @@ if(DEBUG_USE_ASAN AND DEFINED DEBUG_BUILD)
 	CHECK_AND_SET_FLAGS(COMPILER_SUPPORTS_ASAN "-fsanitize=address" DEBUG_FLAGS "-fsanitize=address")
 endif()
 
-if(RELEASE_USE_LTO)
-	set(RELEASE_FLAGS "${RELEASE_FLAGS} -flto")
+if(RELEASE_USE_LTO AND NOT DEFINED BEBUG_BUILD)
```
Akaricchi: * There's a typo here (**B**EBUG_BUILD), and LTO will be enabled for any build type, consequently
* `DEBUG_BUILD` isn't appropriate here in the first place, because it's also set for `RelWithDebInfo`. We want to have LTO available on both `Release` and `RelWithDebInfo` builds. There is a definition of `RELEASE_BUILD` down below, I suggest to move it into the global scope near `DEBUG_BUILD`, and use it in this test.

Akaricchi:
In general, I want to make it very clear that compromising features by using very old versions of software is discouraged, even if supported. That's why I don't like the near-silent fallbacks that require no user intervention. At very least, stuff like that warrants a `WARNING`-level message, certainly not a `STATUS`.

src/CMakeLists.txt
makise-homura: Well, I agree that old cmake should not disable zip support at all, I'll fix it.

About the error: I've made this all while keeping in mind configure script of SDL2_mixer, which automatically disables support for file formats whose libraries do not exist in system (smpeg, ogg+vorbis+vorbisfile, modplug, etc.). But yes, there's opposite approach also exist: Mesa, for example, requires '--enable-gallium-llvm' manually specified in command line, even if there is '--gallium-drivers=radeonsi' exist, which is supposed to use LLVM; so the configure script fails with error, if I specify '--gallium-drivers=radeonsi' without '--enable-gallium-llvm', but adding the latter parameter to configure's command line makes configuration successful.

But is it really required to fail with an error if no suitable libzip found or we can't package files? Maybe warning should be enough?
I think about text like "Your version of CMake is too old, so packaging of data will not be used. Upgrade CMake to at least version 3.3 to enable packaging or specify PACKAGE_DATA=OFF to avoid this warning" and "You have no suitable libzip, so packaging of data will not be used and ZIP support will be disabled. Install at least libzip 1.0 to enable ZIP support or specify DISABLE_ZIP=ON to avoid this warning". Will this be ok?

src/CMakeLists.txt
makise-homura: Yes, looks like it is better way to handle this. I'll fix it too.

Akaricchi:
The version parsing code no longer works correctly at least on CMake 3.8.2 after this change. CMake 3.9.3  is fine.

```
 % ./cmake ../..
== Using MXE wrapper: /data/git/mxe/usr/bin/x86_64-w64-mingw32.static-cmake
     - cmake version 3.8.2
     - warnings for unused CMAKE_POLICY_DEFAULT variables can be ignored
== Using MXE toolchain: /data/git/mxe/usr/x86_64-w64-mingw32.static/share/cmake/mxe-conf.cmake
== Using MXE runresult: /data/git/mxe/usr/share/cmake/modules/TryRunResults.cmake
== Adding "-DCMAKE_BUILD_TYPE=Release"
loading initial cache file /data/git/mxe/usr/share/cmake/modules/TryRunResults.cmake
-- Taisei vv1.1-91-gb747e9f
-- Build configuration: Release
-- Checking for module 'libzip>=1.0'
--   Found libzip, version 1.1.3
-- Configuring done
-- Generating done
-- Build files have been written to: /home/akari/git/taisei/winbuild/x64-static
```

Note the double v. The leading v is not stripped correctly from the git tag. This causes compile errors.

PR #89, opened 2017-10-08
revert pull request #86

Akaricchi:
cmake compatibility issues


Akaricchi:
Note to self: never use the "revert" button on github.

PR #90, opened 2017-10-10
Another attempt to fix things which were in PR #86

makise-homura:
I hope there would not be any bugs with MXE and/or something else...


Akaricchi:
Merged it, but also made some changes of my own in 63d29074f66e453f6bbec5fe8d6abaff4433f235; in particular, I've reversed the option and renamed it to `USE_ZIP` (also done the same for `NO_AUDIO` - it's now `USE_AUDIO`).

makise-homura:
Thanks. I even have some thoughts that it would be better if CMAKE options were one of the four types: disabling only (i.e. there is no ENABLE_SOMETHING option, only DISABLE_SOMETHING), enabling only (same with ENABLE and DISABLE exchanged) default-on (i.e. there is ENABLE option for everything enabled by default, and DISABLE otherwise, so they all are ON), or default-off (DISABLE for everything enabled, and ENABLE otherwise, so they all are OFF). But I considered in not so important, but you already have done that (changed them to 'enabling' type) in mentioned commit. That's nice.

PR #91, opened 2017-10-17
Create GameExplanation.txt

joemaro:
proposing to ship this with the game download, because i had a hard time to figure out anything at the start. i've never played any of the Touhou games and i think you also might want to appeal to "lazy" new people that don't bother asking you this on IRC :)

On Linux, having this as a man-page would be the ideal, but i don't know how to integrate that, so maybe just shipping the text file with the gamedata is a good start.

### DIFF
--- a/GameExplanation.txt
+++ b/GameExplanation.txt
@@ -1,0 +1,30 @@
+Taisei Project
+
+Game Explanation:
+
+* Avoid being shot by the bullets
+
+* There are 3 buttons:
+    * Shoot - Z     - shoots your weapon depending on your current weapon power
+    * Bomb  - X     - kills everything on the screen for some time
+    * Focus - Shift - described further down
+
+* Labels on the right are:
+    * Player = how many lifes you have (only full stars count as a life)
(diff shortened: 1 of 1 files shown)


laochailan:
Thanks, it’s a really good idea :D we didn’t notice there was a demand for this. Will merge it as soon as I have time :)

makise-homura:
I think, what also would be good to describe in this document:

- how do bombs work (it may be intuitively clear for those who played shmup-like games before, but speaking of me, touhou games was first with this mechanic, so I had to find this out myself; if we focus on even random player, it may be useful).
- grazing mechanism
- meaning of points (why it is better to collect more points: additional bomb/life fragments do spawn)
- boss fighting: generic attacks and spell cards, extra spell, health bar, stars meaning remained number of attacks, time limit, 'enemy' pointer at bottom of the screen.

Well, it may be a large list of details, but you may think about it.

laochailan:
I merged and edited the manual and added some of @makise-homura suggestions. :)

Thanks for your contribution @joemaro !

PR #98, opened 2017-11-17
Make the game more cross-compiling friendly

pzmarzly:
TODO - test whether new `vfs_syspath_query` works (it should, according to [SDL docs](https://wiki.libsdl.org/SDL_RWFromFile))

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,23 @@ $prefix/data/
 
 `RELATIVE` is always set when building for Windows or macOS.
 
+## Cross-compiling on Linux for Windows
+
+Get MinGW equivalents of listed dependencies, then run:
+
+```
+mkdir build
+cd build
+mkdir /tmp/taisei-build
+x86_64-w64-mingw32-cmake -DCMAKE_SYSTEM_NAME=Windows -DCMAKE_C_COMPILER=/usr/bin/x86_64-w64-mingw32-gcc -DCMAKE_INSTALL_PREFIX=$output-folder -DRELATIVE=TRUE -DCMAKE_BUILD_TYPE=Release ..
+make
+make install
+```
+
+x86_64-w64-mingw32-gcc may be in different directory on your machine.
+
+This will generate portable build of Taisei in $output-folder.
+
 ## Where are my replays, screenshots and settings?
 
 Taisei stores all data in a platform-specific directory:
--- a/README.md
+++ b/README.md
@@ -51,9 +51,9 @@ make
 make install
 ```
 
-x86_64-w64-mingw32-gcc may be in different directory on your machine.
+`x86_64-w64-mingw32-gcc` may be in different directory on your machine.
 
-This will generate portable build of Taisei in $output-folder.
+This will generate portable build of Taisei in `$output-folder`.
 
 ## Where are my replays, screenshots and settings?
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -145,15 +145,15 @@ if(WIN32)
 	)
 endif()
 
-if(POSIX)
-	set(SRCs ${SRCs}
-		vfs/syspath_posix.c
-		rwops/rwops_pipe_posix.c
-	)
-elseif(WIN32)
+if(WIN32)
 	set(SRCs ${SRCs}
 		vfs/syspath_win32.c
 		rwops/rwops_pipe_null.c
+	)
+elseif(POSIX)
+	set(SRCs ${SRCs}
+		vfs/syspath_posix.c
+		rwops/rwops_pipe_posix.c
 	)
 else()
 	set(SRCs ${SRCs}
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -285,7 +285,7 @@ if(${WIN32})
 	add_definitions(-D__USE_MINGW_ANSI_STDIO)
 	set(LIBs ${LIBs} -ldxguid -lwinmm)
 
-	string(REPLACE "gcc" "windres" CMAKE_RC_COMPILER_INIT ${CMAKE_C_COMPILER})
+	string(REPLACE "gcc" "windres" CMAKE_RC_COMPILER ${CMAKE_C_COMPILER})
 	enable_language(RC)
 	set(CMAKE_RC_COMPILE_OBJECT "<CMAKE_RC_COMPILER> -O coff -o <OBJECT> <SOURCE>")
 
--- a/src/main.c
+++ b/src/main.c
@@ -6,6 +6,8 @@
  * Copyright (c) 2012-2017, Andrei Alexeyev <akari@alienslab.net>.
  */
 
+#define SDL_MAIN_HANDLED
+
 #include <locale.h>
 
 #include "global.h"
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -38,6 +38,23 @@ $prefix/data/
...
 
+## Cross-compiling on Linux for Windows
+
+Get MinGW equivalents of listed dependencies, then run:
+
+```
+mkdir build
+cd build
+mkdir /tmp/taisei-build
+x86_64-w64-mingw32-cmake -DCMAKE_SYSTEM_NAME=Windows -DCMAKE_C_COMPILER=/usr/bin/x86_64-w64-mingw32-gcc -DCMAKE_INSTALL_PREFIX=$output-folder -DRELATIVE=TRUE -DCMAKE_BUILD_TYPE=Release ..
```
lachs0r: The right thing to do here is to use a toolchain file, like the one generated here:
https://github.com/lachs0r/mingw-w64-cmake/blob/master/toolchain.cmake.in

src/CMakeLists.txt
```diff
@@ -145,16 +145,16 @@ if(WIN32)
 	)
 endif()
 
-if(POSIX)
```
lachs0r: Actually, projects like midipix aim to provide POSIX syscalls on Windows using the native NT APIs.

The right thing to do here is to have actual checks for that.

src/CMakeLists.txt
```diff
@@ -285,7 +285,7 @@ if(${WIN32})
 	add_definitions(-D__USE_MINGW_ANSI_STDIO)
 	set(LIBs ${LIBs} -ldxguid -lwinmm)
 
-	string(REPLACE "gcc" "windres" CMAKE_RC_COMPILER_INIT ${CMAKE_C_COMPILER})
```
lachs0r: I’m really not sure what that’s doing here anyway. Neither should be required, and it worked fine with my toolchain file.

lachs0r:
I’m in the process of porting the build system to Meson, anyway (see the meson branch). Cross compiles already work, but I still need to polish a few things and add NSIS installer and macOS app bundle targets.

pzmarzly:
Well, I have done those commits because my friend wanted to get latest build for Windows (and managing libraries on Windows is PITA, managing anything on Windows is PITA tbh). So feel free to do whatever you want with this code, I'm leaving it as it is.

I think 7b6a6d2c8d63c0e860b93212372fcfe0a08ea8e6 and 86e1fcba7f8b796a3d8eee7742cc2b2e22049670 can be merged.

By the way, how are the Windows [releases](https://github.com/taisei-project/taisei/releases) statically linked?

laochailan:
Actually we distribute Debug builds for our windows testers regularly. @Akaricchi does those (and all the windows builds). Most of the development/testing discussion is going on on irc/discord so you can drop by there if you are interested. Always got an open ear for you and your friends gameplay feedback as well :)

src/CMakeLists.txt
Akaricchi: The `POSIX` variable is only true if `_POSIX_VERSION` is defined in `unistd.h`. If this change was actually necessary for you to produce a working build, then there's a good chance your toolchain is broken in some subtle way (maybe looking for system headers when cross-compiling?)

@lachs0r, is it not reasonable to assume that POSIX syscalls are available if we have `_POSIX_VERSION`?

Akaricchi:
@pzmarzly, I use the [MXE toolchain](http://mxe.cc/) with their `*-w64-mingw32.static` targets to cross-compile the windows builds. Unfortunately, some of our dependencies require patches to link properly, which you can find at the [MXE git repo](https://github.com/mxe/mxe).

src/vfs/syspath_win32.c
```diff
@@ -64,10 +63,12 @@ static void _vfs_set_error_win32(const char *file, int line) {
 static VFSInfo vfs_syspath_query(VFSNode *node) {
     VFSInfo i = {0};
 
-    if(!PathFileExists(node->_wpath_)) {
+    SDL_RWops *rwops = SDL_RWFromFile(node->_path_, "rb");
```
Akaricchi: Maybe there's a better option? Opening a file may fail due to many reasons, even though the file actually exists.

src/main.c
```diff
@@ -6,6 +6,8 @@
  * Copyright (c) 2012-2017, Andrei Alexeyev <akari@alienslab.net>.
  */
 
+#define SDL_MAIN_HANDLED
```
Akaricchi: Could you elaborate on this one?

Comment from SDL_main.h:
```
#ifndef SDL_MAIN_HANDLED
#if defined(__WIN32__)
/* On Windows SDL provides WinMain(), which parses the command line and passes
   the arguments to your main function.

   If you provide your own WinMain(), you may define SDL_MAIN_HANDLED
 */
#define SDL_MAIN_AVAILABLE
```

We don't provide our own WinMain, so I'm not sure why would this be needed or how it would even work.

pzmarzly: I was getting [this error](https://stackoverflow.com/a/25089610/5108318) (mingw inserts its own WinMain which loads main).

src/vfs/syspath_win32.c
pzmarzly: I was following [this thread](https://stackoverflow.com/questions/2903360/how-to-check-whether-a-file-exists-in-c-drive-using-vc) - I couldn't get shlwapi to link, stat wouldn't be available for VC++ users, fopen is marked as deprecated by SDL2 which suggests using its own functions.

src/main.c
Akaricchi: Interesting. We use the mingw-w64 fork instead of plain mingw, so I'm guessing that one doesn't do it. I wonder if there's some way to make it work on both without user intervention. @lachs0r, any idea?

src/vfs/syspath_win32.c
Akaricchi: I've just pushed 85c4fb8, which fixes the linking error for me. Can you please check if that works for you as well?

I'm pretty sure that VC++ users won't be able to compile Taisei anyway without some drastic changes. Taisei is written in (mostly standard) C11; as far as I know, VC++ doesn't have good support for it. Our build system also currently expects a GCC or clang -compatible compiler frontend, due to all the hardcoded flags.

That said, stat would most likely not work with non-ASCII paths, anyway.

pzmarzly: I am too lazy to use toolchain file, so I'm running the same flags as before. But now, with fresh CMakeCache.txt I'm hitting new kind of error altogether...
```
x86_64-w64-mingw32-gcc: error: unrecognized command line option ‘-rdynamic’
```

I'm too tired of trying to make cmake work, I will rather wait for meson before further testing.

By the way, I found out that I have unistd.h in mingw headers because of Arch Linux `mingw-w64-headers` package, which is required by `mingw-w64-gcc`.

Akaricchi: I can sympathize, CMake *is* a world of misery and pain. That's why we're moving to Meson in the first place. But hey, at least it's not autohell ;)

`unistd.h` is expected to be in the mingw headers, it shouldn't define `_POSIX_VERSION` though.

lachs0r:
Closing as obsolete due to PR #100.

PR #99, opened 2017-11-17
Remove FL Studio metadata from BGM files

pzmarzly:
Done with

```
find . -type f \( -name "*.ogg" \) -exec vorbiscomment -w -c /dev/null "{}" \;
```


Akaricchi:
Thanks for bringing this to my attention. However, I'm about to push a local branch that rewrites all of the BGMs with properly normalized versions, so we can not merge this. I'll make sure the new files contain no extraneous metadata though.

PR #101, opened 2017-12-25
Reimu

TaiSanta:
Merii Kurisumasu Akari :christmas_tree:


resources/gfx/reimu.ani
```diff
@@ -0,0 +1,3 @@
+rows = 2
```
lachs0r: Test comment for IRC bot.

PR #114, opened 2018-02-12
macos-install-dylibs: ensure dylibs are user-writeable

mistydemeo:
This fixes a bug @Akaricchi and I were discussing in this Homebrew issue: https://github.com/Homebrew/homebrew-core/pull/23341

When the dylibs are copied by `macos-install-dylibs`, it's important to make sure that the copied files are user-writeable; the original files might not be `+x`, and those permissions are carried through when they're copied into the app bundle. If the copied dylibs aren't writeable by the process installing taisei, then `install_name_tool` will fail.

### DIFF
--- a/scripts/macos-install-dylibs.py
+++ b/scripts/macos-install-dylibs.py
@@ -9,7 +9,7 @@ import sys
 import subprocess
 import itertools
 import shutil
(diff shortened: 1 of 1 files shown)


Akaricchi:
Thanks!

PR #118, opened 2018-04-17
[WIP] Projectile prototypes system

Akaricchi:
Or "templates", if you will. For now, this is mainly intended to decouple projectile hitbox sizes from the actual sprites. Additionally, the `timeout` and `timeout_linear` rules have been removed, and replaced with an intrinsic `timeout` property.

Current issues:
* [x] ~~v1.2 replays desync. This patch is not supposed to alter gameplay.~~ *Infeasible to fix at the moment*
* [x] ~~The stage1 background is broken for some strange reason (reeds are missing).~~ *Actually a master regression*
* [ ] Every `PROJECTILE` call needs to be converted to use the prototypes system. For example, this:
```c
PROJECTILE(
    .sprite = "rice",
    .....
);
```
Becomes this:
```c
PROJECTILE(
    .proto = pp_rice,
    .....
);
```
A temporary workaround is provided to map legacy calls to the correct prototypes, where possible. 

* [x] Some spells are not yet tested.


src/stages/stage3_events.c
```diff
@@ -809,9 +810,9 @@ static int wriggle_rocket_laserbullet(Projectile *p, int time) {
 				.sprite = "blast",
 				.pos = p->pos,
 				.color = c,
-				.rule = timeout,
+				.timeout = 35 - 5 * frand(),
```
laochailan: This looks evil

src/stages/stage6_events.c
```diff
@@ -458,16 +460,15 @@ void elly_newton(Boss *b, int t) {
 		}
 
 		PROJECTILE(
-			.sprite_ptr = apple,
-			.pos = clamp(creal(global.plr.pos) + nfrand() * 64, apple->w*0.5, VIEWPORT_W - apple->w*0.5),
```
laochailan: this is the one you already found i suppose?

src/stages/stage3_events.c
Akaricchi: Exactly is evil as the code it replaces, it's a direct adaptation. I've just moved the timeout from args to the new property, exactly the same as in many other places. This is within a `PARTICLE` call, though, so it ought to not desync anything.

src/stages/stage6_events.c
Akaricchi: Indeed. It's unfortunate that we shipped v1.2 with this.

Akaricchi:
Whatever, ship it!

PR #119, opened 2018-05-09
Post-battle revision for Marisa

Lyricana:
Hopefully I'm doing this formatting correctly.


Akaricchi:
The syntax is fine. The only issues are the double-spaces between the sentences, and the apostrophes are probably not the right ones (I don't know which ones to use, either). I'll just have @laochailan deal with that.

The text itself looks good too. The only thing that bugs me a bit is wriggle's last line; "I'll win anyway as long as you lose" sounds rather goofy. I've got no suggestions for now, though.

PR #120, opened 2018-05-09
Post-battle revision for Reimu

Akaricchi:


src/dialog/reimu.c
```diff
@@ -64,11 +71,15 @@ void dialog_reimu_stage4(Dialog *d) {
 
 void dialog_reimu_stage4_post(Dialog *d) {
 	M(Left, "See, I don’t scare easily, so that didn’t work. You should have just told me your name.");
+	M(Right, "I didn't think it was possible, but you're so much stronger than before!");
+	M(Right, "Can't you remember that I'm Kurumi?");
```
Akaricchi: Maybe something like "Can you really not remember me? It's me, Kurumi!" would be better?

src/dialog/reimu.c
```diff
@@ -64,11 +71,15 @@ void dialog_reimu_stage4(Dialog *d) {
...
+	M(Left, "You don't seem like the sort that is worth remembering after several years.");
+	M(Right, "Waah, so mean!  You were way cuter back then!");
 }
```
Akaricchi: Whoah, sick buuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuurn!

That really is kind of mean though. Makes Reimu look like an asshole. I'm not sure how I feel about this :thinking:

src/dialog/reimu.c
```diff
@@ -64,11 +71,15 @@ void dialog_reimu_stage4(Dialog *d) {
...
 
 void dialog_reimu_stage5(Dialog *d) {
 	M(Left, "I didn’t expect someone actually respectable like you to cause me trouble for no reason.");
-	M(Right, "You may be the Hakurei shrine maiden, but I’m afraid that you don’t have permission from Heaven to be here.");
+	M(Right, "You may be the Hakurei Shrine Maiden, but I’m afraid that you don’t have permission from Heaven to be here.");
 	M(Left, "What, are you saying that the gods suddenly don’t think I’m fit for duty anymore?");
 	M(Right, "No, that’s not it. We already have a more suitable candidate from our concerned Celestials, and she is due to arrive instead.");
```
Akaricchi: Not actually related to the changes, I'm just curious: who might that candidate be?

src/dialog/reimu.c
```diff
@@ -83,6 +94,9 @@ void dialog_reimu_stage5_mid(Dialog *d) {
 
 void dialog_reimu_stage5_post(Dialog *d) {
 	M(Left, "See? I’ve been solving incidents since the beginning. There is absolutely no one more qualified than me.");
+	M(Right, "You're still as unruly as ever.  But you passed, and you certainly hold great power.");
+	M(Right, "I'll let my superiors know that the Hakurei Shrine Maiden arrived to clear up this issue.");
+	M(Left, "Right, and I'll make sure it won't take too long.  Tell them to hold tight.");
```
Akaricchi: ... and donate to my shrine? :smile:

src/dialog/reimu.c
Lyricana: The only really wrong thing about it is high informality, but like I told lao, I tend to low-key reduce it when characters are flustered or tired, making them take more verbal shortcuts in their annoyance.  However, like I told lao, I can change it to this if you'd like.  It's still fitting, just a little tonally different.

src/dialog/reimu.c
Lyricana: Not really anyone specific since I don't think Tenshi would be a recommendation, but it's my idea that they're sending a competent celestial from her family (like one of the adults) to handle this issue.

src/dialog/reimu.c
Lyricana: I'll leave it up to you guys' deliberation, but canon Reimu is really not afraid to hand out the stingers, and even told both Alice and Yuuka at this point that they're not worth her memory bank.  So this isn't terribly different from what she already tells 2hus she doesn't care about, haha.

Akaricchi: Hm, I've thought some more about it. I think what actually bothers me here is the one-sidedness of this conversation. The way I see it, all that Kurumi does is praise Reimu's superpowers and complain about being forgotten. It doesn't make for an intimidating vampire image; it makes Kurumi look like a weakling and Reimu look like an asshole for being mean to her.

Maybe one way to fix this might be buffing Kurumi's self-respect and arrogance here. Make her visibly annoyed by having lost the battle and by Reimu's rude remark, and let them have a little sassy back-and-forth perhaps. The aforementioned Alice and Yuuka don't strike me as the kind to go "waah" after being insulted, maybe that's why it works.

I don't know how well this lines up with Kurumi's canon characterization or your vision of it, though, so take that with a massive grain of salt. You don't have to go in this direction if you don't think it's a good idea, of course.

src/dialog/reimu.c
```diff
@@ -96,7 +96,7 @@ void dialog_reimu_stage5_post(Dialog *d) {
 	M(Left, "See? I’ve been solving incidents since the beginning. There is absolutely no one more qualified than me.");
 	M(Right, "You're still as unruly as ever.  But you passed, and you certainly hold great power.");
 	M(Right, "I'll let my superiors know that the Hakurei Shrine Maiden arrived to clear up this issue.");
-	M(Left, "Right, and I'll make sure it won't take too long.  Tell them to hold tight.");
+	M(Left, "Right, and I'll make sure it won't take too long.  Tell them to hold tight, and maybe slip me some yen later.");
```
Akaricchi: Gud change :+1:

src/dialog/reimu.c
Akaricchi: Something like "Can't you remember? I'm Kurumi!" then, maybe? I'll leave it up to you though.

PR #122, opened 2018-05-10
Improve (totally rewrite) projectile collision detection and graze mechanics

Akaricchi:
* Bullets are now *actually* ellipses.
* Player and bullet trajectory is lerped to mitigate phasing through bullets on high velocities
* Some pattern-specific bugs introduced with projectile prototypes fixed.
* You can graze a bullet multiple times now. Each successive graze per bullet reduces the grazing area for that bullet and gives more points.
* Collision areas adjusted for some projectiles.


src/projectile.c
```diff
@@ -159,6 +159,15 @@ void projectile_set_prototype(Projectile *p, ProjPrototype *proto) {
 	p->proto = proto;
 }
 
+complex projectile_graze_size(Projectile *p) {
+	if(p->type == EnemyProj && !(p->flags & PFLAG_NOGRAZE) && p->graze_counter < 5) {
+		complex s = (p->size * 420 /* graze it */) / (2 * p->graze_counter + 1);
+		return sqrt(creal(s)) + sqrt(cimag(s)) * I;
```
laochailan: This does not keep the aspect ratio between w/h equal right? making it sqrt(w/h) in the end which will be closer to a circle.

Alternatively `return sqrt(cabs(s))* (s/cabs(s))` or shorter `return s/sqrt(cabs(s))` this won’t change the aspect ratio because it’s just dividing both by a number. but it should be checked tha

src/util.c
```diff
@@ -344,6 +344,74 @@ float sanitize_scale(float scale) {
...
+	complex m, v;
+	double projection, lv, lm, distance;
+
+	m = seg.b - seg.a; // vector pointing along the line
+	v = seg.a - c.origin; // vector from circle to point A
+
+	lv = cabs(v);
+	lm = cabs(m);
+
+	if(lm == 0) {
```
laochailan: This thing prevents it to work for the case of zero segment width. the mystery check should go before this!

src/util.h
```diff
@@ -100,6 +88,41 @@ typedef struct IntRect {
...
+typedef struct Circle {
+	complex origin;
+	double radius;
+} Circle;
+
+typedef struct Rect {
+	complex top_left;
+	complex bottom_right;
+} Rect;
+
```
laochailan: I bet ellipse and circle will never be used outside of collision code.

src/util.h
```diff
@@ -114,6 +137,10 @@ float smooth(float x) attr_const;
 float smoothreclamp(float x, float old_min, float old_max, float new_min, float new_max) attr_const;
 float sanitize_scale(float scale) attr_const;
 
+bool point_in_ellipse(complex p, Ellipse e) attr_const;
+double lineseg_circle_intersect(LineSegment seg, Circle c) attr_const;
+bool lineseg_ellipse_intersect(LineSegment seg, Ellipse e) attr_const;
+
```
laochailan: maybe a collision.c is better?

src/util.h
Akaricchi: Probably, but we already have collision code in more than one place (projectiles and lasers), and we may want something more sophisticated for bosses/enemies later - Yuyuko's huge-ass shootable banner in PCB comes to mind.

src/util.h
Akaricchi: I do plan to refactor/split up `utils.c` soon™, and I agree these should get their own file. I just didn't want to conflate too many things in this pull request, so I put them there temporarily. I think I'd prefer `geometry.c` or something like that, though.

PR #147, opened 2018-09-14
Add option and keybinding to toggle audio mute.

holycleugh:


Akaricchi:
Hello, thanks for the patch! This seems like a useful little feature. However, I do have a few issues with this patch, which I'll outline shortly. I'll be happy to merge it when they've been resolved.

src/audio.h
```diff
@@ -79,3 +79,5 @@ void fade_bgm(double fadetime);
 void resume_bgm(void);
 void save_bgm(void); // XXX: this is broken
 void restore_bgm(void); // XXX: this is broken
+
+void mute_audio_handler(void);
```
Akaricchi: Internal functions should be declared static, and not exposed in the headers. I'm aware that not all code follows this rule, but that just means it should be fixed/refactored.

src/events.c
```diff
@@ -412,6 +412,12 @@ static bool events_handler_key_down(SDL_Event *event, void *arg) {
 		}
 	}
 
+	if (!repeat) {
+		if(scan == config_get_int(CONFIG_KEY_TOGGLE_AUDIO)) {
+			config_set_int(CONFIG_MUTE_AUDIO, !config_get_int(CONFIG_MUTE_AUDIO));
```
Akaricchi: This probably should be moved into [`events_handler_hotkeys`](https://github.com/taisei-project/taisei/blob/8b7a259759e7cdb7f336af3bf894fa2bd6f6c30e/src/events.c#L440).

src/audio_common.c
```diff
@@ -305,6 +305,8 @@ static bool audio_config_updated(SDL_Event *evt, void *arg) {
 			break;
 	}
 
+	mute_audio_handler();
```
Akaricchi: This looks like it works, however, the switch above no longer does anything useful and can be removed entirely. The `mute_audio_handler` code can just be moved into this function as well.

holycleugh:
Good catch, fixed.

PR #157, opened 2019-02-01
Some compatibility fixes for Elbrus compiler (and others based on EDG front end)

makise-homura:
There were 5 issues that produced warnings: stack variable alignment, typedef redefinition, warning on unreachable __builtin_unreachable(), unsigned enums and __attribute__((designated_init)). Fixes were done to achieve warningless release build (though debug still have some -Wtype-limits on asserts due to this unsigned enums stuff).

Tested on following setups:
x86_64, meson 0.45.1, gcc (Ubuntu 7.3.0-27ubuntu1~18.04) 7.3.0
x86_64, meson 0.45.1, clang version 6.0.0-1ubuntu2 (tags/RELEASE_600/final) Target: x86_64-pc-linux-gnu
Elbrus-8C, meson 0.48.999, lcc:1.23.10:Jun-11-2018:e2k-v4-linux gcc (GCC) 5.5.0 compatible

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -48,7 +48,7 @@ taisei_c_args = cc.get_supported_arguments(
     '-Winit-self',
     '-Wlogical-op',
     '-Wmissing-prototypes',
-    '-Wno-typedef-redefinition'
+    '-Wno-typedef-redefinition',
     '-Wno-long-long',
     '-Wnull-dereference',
     '-Wparentheses',
--- a/meson.build
+++ b/meson.build
@@ -48,6 +48,7 @@ taisei_c_args = cc.get_supported_arguments(
     '-Winit-self',
     '-Wlogical-op',
     '-Wmissing-prototypes',
+    '-Wno-typedef-redefinition'
     '-Wno-long-long',
     '-Wnull-dereference',
     '-Wparentheses',
--- a/meson.build
+++ b/meson.build
@@ -56,20 +56,10 @@ taisei_c_args = cc.get_supported_arguments(
     '-Wsometimes-uninitialized',
     '-Wstrict-prototypes',
     '-Wtype-limits',
+    '-Wunreachable-code',
     '-Wunneeded-internal-declaration',
     '-Wunreachable-code-loop-increment',
 )
-
-if cc.compiles('''
-int main() {
-    return 0;
-    __builtin_unreachable();
-}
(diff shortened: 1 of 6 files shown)


src/renderer/common/sprite_batch.c
```diff
@@ -183,7 +183,7 @@ void r_flush_sprites(void) {
 }
 
 static void _r_sprite_batch_add(Sprite *spr, const SpriteParams *params, SDL_RWops *stream) {
-	SpriteAttribs alignas(32) attribs;
+	SpriteAttribs alignas(16) attribs;
```
Akaricchi: `SpriteAttribs` is already aligned to 16 bytes (because of the mat4 members), so there's no need to force alignment here.

src/util/meson.build
```diff
@@ -23,6 +23,16 @@ if is_developer_build
     util_src += files('debug.c')
 endif
 
+use_designated_init = cc.compiles('''
+struct { ; } __attribute__((designated_init)) x;
```
Akaricchi: Empty structs are a GNU extension, you should add a dummy member.

meson.build
```diff
@@ -56,10 +57,20 @@ taisei_c_args = cc.get_supported_arguments(
...
+int main() {
+    return 0;
+    __builtin_unreachable();
+}
+''', name : 'unreachable __builtin_unreachable() with -Werror', args : [ '-Wunreachable-code', '-Werror' ] )
+    taisei_c_args += cc.get_supported_arguments('-Wunreachable-code')
+else
+    warning('-Wunreachable-code will not be used due to incorrect handling of __builtin_unreachable')
+endif
+
```
Akaricchi: I'm not sure this check is necessary. I think it's fine to let buggy compilers spit bogus warnings, so long as they are harmless. Does the Elbrus compiler suffer from this issue as well?

makise-homura: Yes, it was caused by me misunderstanding cause of some warnings, and after commit 36493d3, it is no more needed.

src/renderer/common/sprite_batch.c
makise-homura: Fixed in d38a434.

src/util/meson.build
makise-homura: Fixed in a1b4729.

Akaricchi:
Thank you!

PR #168, opened 2019-06-11
always update gamepad device list at least once

ailin-nemui:
I had this issue with the Nintendo GameCube adapter on Linux, that it would show "No devices available" and controller input wouldn't work.

@Akaricchi pointed me to this quick fix. There may be better solution, or further understanding why SDL events don't work on my system for this particular controller (SDL 2.0.9 Linux openSUSE 15.0 x64)


Akaricchi:
I think this is fine; the worst that could happen is the device list getting updated twice in a row at init. Thanks for the finding and the PR.

PR #169, opened 2019-06-19
Update gamepads unconditionally, once

ailin-nemui:
The commit of #168 was incomplete in that it did not enable the device by setting active_dev_num


Akaricchi:
Actually I think it makes sense to just do `gamepad.update_needed = true;` there instead of calling the function directly. It'll be called on the next frame then, and if any hotplug events arrive before that happens, they won't cause the list to be needlessly updated twice.

PR #170, opened 2019-06-23
New character art by Afensorm + other visual tweaks

Akaricchi:
Mostly ready; got to fix the dialog portrait draw order for the overlap case and awaiting tweaks to some of the earlier art.


Akaricchi:
Oh yeah, I forgot about Wriggle's glowing antennae: I still haven't made it look more like a proper glow effect; that should be fixed before merge too.

PR #171, opened 2019-06-25
Resource system refactoring

Akaricchi:
This is a backport from my work on the lua branch.


Akaricchi:
Closing this as stale and outdated. There are some good ideas here that are worth porting eventually (refcounting, removal of permanent/transient dichotomy in favor of ref groups, heuristic delayed unloading), but the opaque ref handles didn't quite work out.

PR #173, opened 2019-07-31
Switch homebrew port

p-sam:
* Build scripts and specific readme with instructions in the ./switch folder
* Uses libnx, devkitA64 and portlibs
* Ifdef platform specific code where needed
* Controllers force-enabled on start and option hidden (only targetting Switch)
* Player name option is hidden (only targetting Switch)


scripts/pack.py
```diff
@@ -40,7 +39,7 @@ def pack(args):
 
     with ZipFile(str(args.output), 'w', ZIP_DEFLATED, **zkwargs) as zf:
         for path in sorted(args.directory.glob('**/*')):
-            if path.name[0] == '.' or path.name == 'meson.build':
+            if (path.name[0] == '.' or (args.exclude is not None and next((True for x in args.exclude if path.match(x)), False))):
```
Akaricchi: Well that's a mouthful. I'd advice defaulting `args.exclude` to `[]` (`default=[]` at `parser.add_argument` should do it), and using something like this for the condition: `any(path.match(x) for x in args.exclude)`

src/meson.build
```diff
@@ -248,6 +248,53 @@ if host_machine.system() == 'emscripten'
...
     )
+elif host_machine.system() == 'nx'
+    taisei_elf_name = '@0@.elf'.format(taisei_basename)
+    taisei_elf = executable(taisei_elf_name, taisei_src, version_deps,
+        dependencies : taisei_deps,
+        c_args : taisei_c_args,
+        c_pch : 'pch/taisei_pch.h',
+        install : true,
+        install_dir : bindir,
+        override_options: ['strip=true'],
```
Akaricchi: Wasn't this supposed to be `false`?

meson.build
```diff
@@ -117,7 +117,7 @@ if sm_check.stderr() != ''
     warning('Submodule check completed with errors:\n@0@'.format(sm_check.stderr()))
 endif
 
-static = get_option('static') or (host_machine.system() == 'emscripten')
+static = get_option('static') or (host_machine.system() in ['emscripten', 'nx'])
```
Akaricchi: The `x in list` syntax was added in meson 0.49.0; unfortunately we still need to support at least 0.46.0 to some extent. It's ok if the switch port won't build on it, but the file should at least parse.

Changing it to `['emscripten', 'nx'].contains(host_machine.system())` should fix the issue.

src/arch_switch.c
```diff
@@ -0,0 +1,111 @@
...
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <switch/runtime/devices/socket.h>
+#include <switch/runtime/nxlink.h>
+#include <switch/services/applet.h>
+#include <switch/services/fs.h>
+
+#include "taisei.h"
```
Akaricchi: This should go before all other includes for two reasons:

* We use precompiled headers by default, and the pch includes this file early, so that's more consistent.
* We have an manual upkeep script that inserts missing copyright notices, include guards, and `#include "taisei.h"`s. It'll get confused if this include is not at the top.

You also don't need any of those standard library includes; `"taisei.h"` already includes them.

src/arch_switch.c
```diff
@@ -0,0 +1,111 @@
...
+			taisei_quit();
+			break;
+
+		default:
+			break;
+	}
+}
+
+attr_used
+void userAppInit() {
```
Akaricchi: Functions that take no parameters should be `(void)`; `()` means unspecified parameters.

src/arch_switch.c
```diff
@@ -0,0 +1,111 @@
...
+	NX_SETENV("NV50_PROG_DEBUG", "1");
+	NX_SETENV("NV50_PROG_CHIPSET", "0x120");
+#else
+	// disable error checking and save CPU time
+	NX_SETENV("MESA_NO_ERROR", "1");
+#endif
+}
+
+attr_used
+void userAppExit() {
```
Akaricchi: Same

src/arch_switch.c
```diff
@@ -0,0 +1,111 @@
...
+}
+
+attr_used
+void userAppInit() {
+	socketInitializeDefault();
+	appletLockExit();
+	appletHook(&g_hookCookie, onAppletHook, NULL);
+
+#ifdef DEBUG
+	dup2(1, 2);
```
Akaricchi: Not a complaint, just curious: what is this for?

src/gamepad.h
```diff
@@ -44,10 +44,17 @@ typedef enum GamepadEmulatedButton {
 typedef enum GamepadButton {
 	// must match SDL_GameControllerButton
 	GAMEPAD_BUTTON_INVALID = -1,
+#ifdef __SWITCH__
+	GAMEPAD_BUTTON_B,
+	GAMEPAD_BUTTON_A,
+	GAMEPAD_BUTTON_Y,
+	GAMEPAD_BUTTON_X,
+#else
```
Akaricchi: I'm not sure, but it looks like this would change the relative physical layout of the default controls (as compared to an xbox-like gamepad), wouldn't it? If so, maybe it'd be better to keep the internal IDs as is, and swap the user-visible names instead?

But I suppose if the default controls are still reasonably comfortable, it doesn't matter that much.

src/menu/options.c
```diff
@@ -662,9 +662,11 @@ static MenuData* create_options_menu_gamepad(MenuData *parent) {
 
 	OptionBinding *b;
 
+#ifndef __SWITCH__
 	add_menu_entry(m, "Enable Gamepad/Joystick support", do_nothing,
 		b = bind_option(CONFIG_GAMEPAD_ENABLED, bind_common_onoff_get, bind_common_onoff_set)
 	);	bind_onoff(b);
+#endif
 
 	add_menu_entry(m, "Device", do_nothing,
```
Akaricchi: I'm curious, does this actually work? If I understand right, you can attach multiple controllers to the switch; do you know if they show up here correctly?

p-sam: What happens is that SDL lists 8 controllers no matter what, and then whether you receive inputs or not on any of them will depend of controller actually attached and their order. So what you see in the menus is "Switch controller #1" regardless of the type (pro controller, single joycon, joycon pair, etc)
Btw we could also in the future add an option that would split the joycons, allowing to use a single joycon to play, but i'm not sure if it's worth the trouble tho. (the pair is seen as a single controller by default, and we can talk to the hid service to merge or split any)

Refs:
SDL joystick init: https://github.com/devkitPro/SDL/blob/switch-sdl2/src/joystick/switch/SDL_sysjoystick.c#L74
Hid init: https://github.com/switchbrew/libnx/blob/01d648e7ed211d1e57b6026a1f9fc15cd5c7763a/nx/source/services/hid.c#L106

src/arch_switch.c
p-sam: When doing dev, it's pretty useful to have nxlink and send the NRO that way so that the log shows live on the console, streamed from the device. However, libEGL, and mesa debug logs are outputted on stderr, so it simply redirects stderr to stdout, so that we catch those too.

src/meson.build
p-sam: Most other HB projects also ship the .elf next to the nro by default, and while we don't usually bundle it into release, keeping them around is useful, unlike the nacp

src/gamepad.h
p-sam: They seemed fine but I also did not play more than 2 stages in a row. I'm okay with changing the names instead, it would help the theoretical use case where someone imports its config from PC -> would keep the layout

src/meson.build
Akaricchi: I'm confused; are you talking about the `install`? Because I was talking about the `strip=true`.

src/menu/options.c
Akaricchi: Hm well, that's suboptimal. In that case maybe this option should be hidden (and defaulted to "any device") too, unless it is actually important to actually filter unwanted devices out.

I'm not seeing the benefit of splitting the joycons. It could be useful later if ever add local multiplayer though, but that's probably not happening in the nearest future.

src/gamepad.h
p-sam: Actually I forgot, changing names would invert Back/Accept in menu actions, so that's why I changed only the enum order, which seemed cleaner. I'd prefer letting this than having enums different in the code than what's actually displayed on the menu, and on the controller, and having to add more ifdef to menu code. Tell me what you think of it

Akaricchi: Yeah, in that case I think this is fine.

PR #181, opened 2019-12-08
WIP - new story dialogue

StarWitch:
Hi all. I've taken on rewriting the primary dialogue for Reimu/Marisa/Youmu.

Currently, stage 1 is complete. I'll be updating the PR when more of it is written, and that way it'll be easier to comment on commit-by-commit than one huge text dump.


Akaricchi:
Thanks again for contributing!

Can you please use tabs instead of spaces for indentation? Unfortunately that is our convention. If you use an editor that supports [editorconfig](https://editorconfig.org/), it should pick up the settings automatically.

src/dialog/reimu.c
```diff
@@ -15,21 +15,25 @@
...
+        RIGHT("No way! Prepare to get frozen!");
 }
 
 DIALOG_SCRIPT(reimu_stage1_post_boss) {
-	LEFT_FACE(happy);
-	LEFT("FIXME: Removed due to issue #179");
-	// RIGHT("FIXME: Removed due to issue #179");
-	LEFT_FACE(normal);
-	LEFT("FIXME: Removed due to issue #179");
+        LEFT_FACE(amused);
```
Akaricchi: Pretty sure this face doesn't exist ([here are all of them](https://github.com/taisei-project/taisei/tree/master/atlas/portraits/dialog)).

Doesn't really matter at this stage though; it's fine to completely ignore the faces for now. We can add them later, and I can probably ask @afensorm to draw extra expressions if we end up needing any, too.

StarWitch: Changed that one, yep.

PR #183, opened 2019-12-11
fix macOS compile issue for `timespec_get`

StarWitch:
As seen here (thanks @Akaricchi): https://stackoverflow.com/questions/32803754/what-is-the-darwin-c-level-c-preprocessor-symbol

Also adds some useful tips in the readme for building on Linux/macOS, and adds *.swp files as ignored (common for vim users).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -38,7 +38,7 @@ The build system will try to pull in as many missing dependencies as it can into
 
 On **Linux**, you should use the libraries provided by your distribution's package manager (i.e: apt, yum, etc.). In particular, installing packages like ``libsdl2-dev`` is necessary, especially if your system uses Wayland, since it will pull in the necessary libraries to build for your system. Ensure that you've correctly installed the dependency packages if you run into errors like ``SDL_Error: No available video device``.
 
-On **macOS**, it's the opposite. Let the project pull in as much as possible. Using ``brew`` will *probably not work well,* even for basic dependencies like SDL2. If you run into compile-time issues, try seeing if you have any of those dependencies above installed and remove them. 
+On **macOS**, it's the opposite. You can try to use dependencies from ``brew`` but if anything fails you're better off uninstalling those dependencies and using whatever Meson pulls in at compile-time. 
 
 Build-only dependencies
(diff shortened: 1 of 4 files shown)


meson.build
```diff
@@ -220,6 +220,7 @@ if macos_app_bundle
     datadir = join_paths(bundle_dir, 'Contents', 'Resources')
     bindir = join_paths(bundle_dir, 'Contents', 'MacOS')
     config.set('TAISEI_BUILDCONF_RELATIVE_DATA_PATH', true)
+    config.set('TAISEI_DARWIN_SOURCE', true)
```
Akaricchi: This should be called `TAISEI_BUILDCONF_DARWIN_SOURCE` for consistency if we're doing it this way. It's also under the wrong conditional, because `macos_app_bundle` can be `false` if the `macos_bundle` option is disabled (you get a traditional unix-like directory structure instead of an app bundle in that case).

This should be under an `if host_machine.system() == 'darwin'` clause — that'll cover macOS, iOS, and other Apple systems in case someone decides to make a port. Alternatively, I think it's fine to just check for the `__APPLE__` macro in `util/compat.h` instead.

src/util/compat.h
```diff
@@ -15,6 +15,11 @@
 #define _POSIX_C_SOURCE 200809L
 #endif
 
+// Apple doing weird stuff again
```
Akaricchi: This is probably worth briefly explaining, e.g. "_POSIX_C_SOURCE makes some standard definitions disappear from Apple's headers unless this is defined". Expletives permitted.

README.rst
```diff
@@ -31,6 +31,15 @@ Optional:
 -  SPIRV-Cross >= 2019-03-22 (for OpenGL ES backends)
 -  libshaderc (for OpenGL ES backends)
 
+Platform-Specific Instructions
```
Akaricchi: Could the readme changes please go in a separate PR? I am very much in favor of documenting the dependency subprojects, though.

StarWitch:
Done!

Akaricchi:
Thank you!

PR #184, opened 2019-12-12
macOS/Linux building notes, +*.swp for .gitignore

StarWitch:
Some things I noticed building on macOS and Linux. A section for Windows will be included when I get around to trying to build on Windows too. 

Also adds *.swp for .gitignore, which helps elitist vim users (such as myself).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -48,7 +48,7 @@ The build system will try to pull in as many missing dependencies as it can into
 
 On **Linux**, you should use the libraries provided by your distribution's package manager (i.e: apt, yum, etc.). In particular, installing packages like ``libsdl2-dev`` is necessary, especially if your system uses Wayland, since it will pull in the necessary libraries to build for your system. Ensure that you've correctly installed the dependency packages if you run into errors like ``SDL_Error: No available video device``.
 
-On **macOS**, it's a bit different. Using ``brew`` (from Mac Homebrew), you can safely install: ``libzip``, ``libpng``, ``webp``, ``sdl2``, ``freetype``. What you should **avoid** is installing ``sdl2_mixer`` and ``opusfile``, as they versions they ship don't currently seem compatible with Taisei. The build system will pull those in for you so you can safely ignore them.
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -41,6 +41,15 @@ Optional:
 
 -  docutils (for documentation)
 
+Platform-Specific Instructions
+""""""""""""""""""""""""""""""
+
+The build system will try to pull in as many missing dependencies as it can into the build environment. However, this has a few caveats.
```
Akaricchi: I'm not sure but maybe it's worth mentioning the mechanism used for this (meson subprojects), as well as the fact that you can force it to ignore system libraries or refuse to download subprojects (meson's `--wrap-mode` option).

README.rst
```diff
@@ -41,6 +41,15 @@ Optional:
...
+
+On **Linux**, you should use the libraries provided by your distribution's package manager (i.e: apt, yum, etc.). In particular, installing packages like ``libsdl2-dev`` is necessary, especially if your system uses Wayland, since it will pull in the necessary libraries to build for your system. Ensure that you've correctly installed the dependency packages if you run into errors like ``SDL_Error: No available video device``.
```
Akaricchi: I don't think Wayland is worth mentioning in the system-SDL case; the wayland headers aren't actually required for that (but the runtime libs may be). Maybe it should just recommend to use system libraries with `meson --wrap-mode=nofallback`, unless the user knows what they're doing.

The real caveat with SDL on Linux is that if you build it as a subproject, it will attempt to include support for as much stuff as possible. We can't really make missing libraries an error there, because pretty much everything is optional, and wrapping all of them in subprojects would be a maintenance nightmare.

I'm not sure how to handle this in the readme honestly, perhaps simply listing which libraries enable which functionality in SDL.

Akaricchi:
Hey @StarWitch, sorry for neglecting this PR for a while, I haven't forgotten about it though!

I've left a few comments but I'm not sure what to do with it right now to be honest... writing documentation is hard :(

StarWitch:
Closing this so it can be refreshed later!

PR #189, opened 2020-02-18
sin() broken on values over 10,000 on Intel (shaders)

StarWitch:
Like it says on the tin. This ought to fix that weird checkerboard pattern on the Stage 1 water effect shader, but it's worth checking all other shaders too to make sure their input values are under 10,000, lest Intel's silicon eat our precious danmaku.

To be clear, this is a known, long-standing *hardware* defect that Intel has basically decided not to fix, lest they break backwards compatibility with people's workarounds. 

https://stackoverflow.com/questions/21464664/sinx-only-returns-4-different-values-for-moderately-large-input-on-glsl-fragme

https://randomascii.wordpress.com/2014/10/09/intel-underestimates-error-bounds-by-1-3-quintillion/

image

image


Akaricchi:
Oh wow, interesting find. The second link seems to talk about the CPU instruction (I've only skimmed through it), is it known to apply to the graphics cores as well?

Out of curiosity, where does the 10,000 magic value come from? Or is that an estimate?

I guess we can alias `sin(x)` to `sin(mod(x, 2*pi))` on Intel globally if this becomes a problem elsewhere.

resources/00-taisei.pkgdir/shader/stage1_water.frag.glsl
```diff
@@ -12,8 +12,13 @@ float smoothNoise(vec2 p) {
...
 	f *= f * (3 - f - f);
 
+	// Intel's sin() function breaks down with values over 1e4 (yes, really)
+	// get rid of some bits here w/o sacrificing the nice effect
+	vec4 a = vec4(0, 1, 27, 28) + p.x + p.y * 27;
+	vec4 b = a * mod(1e5, tau);
+
 	// WARNING: Some versions of the Windows AMD driver choke on temp_mat = mat2(temp_vec)
-	vec4 temp_vec = fract(sin(vec4(0, 1, 27, 28) + p.x + p.y * 27) * 1e5);
+	vec4 temp_vec = fract(sin(b));
```
Akaricchi: Wait, now I'm actually confused by this. In the original code:
```glsl
fract(sin(vec4(0, 1, 27, 28) + p.x + p.y * 27) * 1e5)
```
The 1e5 factor is actually outside of the sin's argument... i.e.
```glsl
vec4 sin_arg = vec4(0, 1, 27, 28) + p.x + p.y * 27;
vec4 temp_vec = fract(sin(sin_arg) * 1e5);
```

But in the rewritten form it goes inside the sin. And ` mod(1e5, tau)` is actually a constant (`~3.10583...`). What gives?

PR #190, opened 2020-02-19
OpenGL ES 3.0/ANGLE support on macOS

StarWitch:
This fixes (mostly) everything needed to compile GL ES 3.0 on macOS.

You still need to have `libGLESv2.dylib` and `libEGL.dylib` handy and specified with configuring the environment with Meson, but it should pull those files in and put them in a predictable place within `Taisei.app`. 

I also got rid of some deprecation warnings in `meson.build` files for `build_always`. The replacement functions have been around since before the minimum version to build Taisei (>=0.48) so it should be safe.

Even if it's mostly just a wrapper around OpenGL 4.1 through ANGLE, it's nice to keep our options open. It also runs pretty well. (Who knows - maybe an iOS or Android port for v1.5? (more like v2.0))

### DIFF
--- a/src/renderer/glescommon/gles.c
+++ b/src/renderer/glescommon/gles.c
@@ -17,31 +17,16 @@ void gles_init(RendererBackend *gles_backend, int major, int minor) {
 #ifdef TAISEI_BUILDCONF_HAVE_ANGLE
 	// for loading ANGLE libraries
 	char *basepath = SDL_GetBasePath();
-	size_t basepath_len = strlen(basepath);
-	char buf[basepath_len + 256];
-
-#ifdef __APPLE__
-	const char *libgles = "libGLESv2.dylib";
-	const char *libegl = "libEGL.dylib";
-	// SDL_GetBasePath() returns Contents/Resources
-	// but we need Contents/MacOS, so go up one dir 
-	snprintf(buf, sizeof(buf), "%s../MacOS/ANGLE/", basepath);
-	basepath_len += sizeof("../MacOS/ANGLE");
-#else
-	// for Windows
-	const char *libgles = "libGLESv2.dll";
-	const char *libegl = "libEGL.dll";
-	snprintf(buf, sizeof(buf), "%sANGLE\\", basepath);
-	basepath_len += sizeof("ANGLE");
-#endif
-
-	SDL_free(basepath);
-
-	strlcpy(buf + basepath_len, libgles, sizeof(buf) - basepath_len);
+	char buf[128];
+	
+	// SDL_*_DRIVER are SDL-specific env vars
+	snprintf(buf, sizeof(buf), "%s%s", basepath, TAISEI_BUILDCONF_ANGLE_GLES_PATH);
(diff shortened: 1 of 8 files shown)


src/renderer/glescommon/gles.c
```diff
@@ -14,18 +14,37 @@
...
 void gles_init(RendererBackend *gles_backend, int major, int minor) {
-#ifdef TAISEI_BUILDCONF_HAVE_WINDOWS_ANGLE_FALLBACK
+#ifdef TAISEI_BUILDCONF_HAVE_ANGLE
+	// for loading ANGLE libraries
 	char *basepath = SDL_GetBasePath();
 	size_t basepath_len = strlen(basepath);
-	char buf[basepath_len + 32];
+	char buf[basepath_len + 256];
+
+#ifdef __APPLE__
```
Akaricchi: I think it's best to handle this in Meson rather than ifdeffing it. Have Meson export two new `TAISEI_BUILDCONF_` constants that contain paths to the GLESv2 and EGL libraries, relative to `SDL_GetBasePath()`. Then the C side can be simpler and OS-agnostic.

Note that it's also possible to configure a macOS build that does not produce an app bundle, by disabling the `macos_bundle` meson option. That should be either correctly handled, or made an explicit error with `install_angle`.

src/renderer/meson.build
```diff
@@ -61,8 +61,7 @@ endforeach
 r_macro = ' '.join(r_macro)
 config.set('TAISEI_BUILDCONF_RENDERER_BACKENDS', r_macro)
 config.set_quoted('TAISEI_BUILDCONF_RENDERER_DEFAULT', default_renderer)
-config.set('TAISEI_BUILDCONF_HAVE_WINDOWS_ANGLE_FALLBACK',
-    host_machine.system() == 'windows' and
+config.set('TAISEI_BUILDCONF_HAVE_ANGLE',
     enabled_renderers.contains('gles30') and
```
Akaricchi: This check is probably not needed here, especially considering that one can enable `gles20` without `gles30`. But It may be a good idea to print a warning if `install_angle` is enabled but neither `gles20` nor `gles30` are.

src/renderer/glescommon/gles.c
StarWitch: Fixed relative install paths for macOS and implemented the `TAISEI_BUILDCONF_*` suggestion.

Akaricchi:
The indentation is still broken unfortunately.

(gray arrows are tabs on the screenshot)
screenshot

For meson files (and also Python scripts, and some other things) we use 4 spaces for indentation - I'm sorry for the annoying discrepancy with our C code indentation here, but that ship has sailed.

These settings are specified in our [.editorconfig](https://github.com/taisei-project/taisei/blob/master/.editorconfig), and I recommend [installing an EditorConfig plugin](https://github.com/editorconfig/editorconfig-vim) to make the editor automatically pick them up.

StarWitch:
OK, I've got EditorConfig set up on vim finally. Sorry about that!

PR #191, opened 2020-02-19
Improved documentation

StarWitch:
A work-in-progress. Adds some tips for compiling things, especially on macOS, but also for development in general.


README.rst
```diff
@@ -6,59 +6,43 @@ Taisei
...
 
-You can find the source tarballs at the
-`Releases <https://github.com/taisei-project/taisei/releases>`__ section on
-Github. **Do not** grab Github's auto-generated source archives, those do not
-contain the required submodules. This only applies for versions v1.3 and above.
+::
 
-Latest code from git
-""""""""""""""""""""
+    git clone https://github.com/taisei-project/taisei
```
Akaricchi: It's probably better to include the `--recurse-submodules` option in this command line, and explain its importance below (as well as how to update the submodules).

README.rst
```diff
@@ -6,59 +6,43 @@ Taisei
 Introduction
 ------------
 
-Taisei is an open clone of the Tōhō Project series. Tōhō is a one-man project of
-shoot-em-up games set in an isolated world full of Japanese folklore.
+Taisei Project is an open clone of the Tōhō Project series. Tōhō is a one-man
```
Akaricchi: I know it says "clone" in the original as well, but I think it's a good opportunity to correct that - we should emphasize that it is a fangame, and not an attempt to recreate any of ZUN's games.

It has been unfortunately misconstrued in the past ([here is one of the most painful examples](https://community.chrono.gg/t/taisei-bullet-hell-madness/11065)). That's why our website even links to the wikipedia page on fangames.

README.rst
```diff
@@ -133,74 +89,22 @@ Taisei stores all data in a platform-specific directory:
...
-::
+If you find any other bug not covered in that, feel free to
+`open up an issue <https://github.com/taisei-project/taisei/issues>`__. If you
+do, please give as much information as you can, such as:
 
-    ALSA lib pcm.c:7234:(snd_pcm_recover) underrun occurred
+    1. OS version (i.e: Windows 10, macOS 10.16, Ubuntu Linux 19.10)
+    2. Hardware (i.e: Intel processor, AMD/NVIDIA graphics card, etc)
+    3. Specific problem you're experiencing
+    4. Logs, screenshots, or how to reproduce the the issue
```
Akaricchi: I like the idea, but it's probably better implemented as a github issue template. I don't think it's possible to submit one in a pull request, but I can add it after this lands.

StarWitch: Sure, sounds good!

doc/DEVELOPMENT.rst
```diff
@@ -169,7 +169,7 @@ tools:
 
 ::
 
-    brew install meson cmake pkg-config docutils
+    brew install meson cmake pkg-config docutils imagemagick pygments python3
```
Akaricchi: meson requires python3, so maybe the later is redundant? Unless brew packages it weirdly; I wouldn't know anything about that.

doc/DEVELOPMENT.rst
```diff
@@ -216,3 +216,39 @@ Automatic
...
+
+You'll also need to download (and/or compile) and install
+`Leanify <https://github.com/JayXon/Leanify>`__.
+
+Run the following command from the root directory, once your sprites are
+modified:
+
+::
+
+    scripts/gen-atlas.py atlas/overrides atlas resources/00-taisei.pkgdir/gfx
```
Akaricchi: You shouldn't actually have to run that script manually, Taisei has ninja targets for this with appropriate settings. You can run `ninja gen-atlases` to regenerate all atlases, or `ninja gen-atlas-NAME` to regenerate a specific one.

You can also do `ninja gen-atlases-fast` or `ninja gen-atlas-NAME-fast` to generate atlases quicker, but that should only be used for local testing and not go into commits, as atlases generated this way are larger and less optimized.

doc/DEVELOPMENT.rst
```diff
@@ -216,3 +216,39 @@ Automatic
 
 To do the above steps automatically, run ``scripts/macos-deps-install.sh``.
 
+Graphical Assets
+----------------
+
+Taisei's GFX library is made up of a collection of sprites, shaders, and a few
+3D models. The 3D models are exclusively used for background scenery, while the
```
Akaricchi: Technically the 3D models are not used exclusively for backgrounds - the HUD is a flat 3D model with a cutout for the game's viewport :smile: That was done as a small rendering optimization. But technically nothing prevents you from using models in 2D context in general, they are just a bunch of packed vertex data after all. So perhaps "mostly" or "almost exclusively" is a better way of putting it.

doc/DEVELOPMENT.rst
```diff
@@ -216,3 +216,39 @@ Automatic
...
+danmaku bullets themselves.
+
+To modify the 3D models, you'll need `Blender <https://blender.org>`__, which is
+free and open source. Look for tutorials on YouTube for how to use it. The
+models themselves are located in ``resources/00-taisei.pkgdir/models``.
+
+Music and sound effects are located in ``resources/00-taisei.pkgdir/sfx``.
+
+For sprites, any image editor will do. Sprites are locaed in ``atlas``, in
+``.png`` format. However, to have sprites properly appear in Taisei, you'll
```
Akaricchi: Lossless `.webp` should generally be preferred for all new images (they are significantly smaller); the PNGs in the `atlas/` dir are mostly remains from before the webp switch. We have a script to automatically optimize images, `scripts/optimize-img.sh` and it's whole-directory wrapper, `scripts/optimize-all-img.sh` (these probably should be rewritten in Python eventually).

The script will optimize PNGs with [Leanify](https://github.com/JayXon/Leanify), convert them to WEBPs (with the official `cwebp` tool), and keep whichever is smaller (usually the webp, but there are rare exceptions). It'll also try to losslessly recompress WEBPs. Generally it should be ran on all newly added or edited images, except atlases (the atlas generator optimizes them already).

doc/DEVELOPMENT.rst
```diff
@@ -249,6 +249,6 @@ modified:
 
 ::
 
-    scripts/gen-atlas.py atlas/overrides atlas resources/00-taisei.pkgdir/gfx
+    ninja gen-atlas-common_ui
```
Akaricchi: Ah well, I see you edited this before I submitted my earlier comment. It's still not quite right though, so that comment still applies - this command will regenerate just the `common_ui` atlas specifically.

doc/DEVELOPMENT.rst
StarWitch: Updated to use the `ninja` command!

doc/DEVELOPMENT.rst
```diff
@@ -0,0 +1,266 @@
...
+
+    brew install freetype2 libzip opusfile libvorbis webp sdl2
+
+
+You'll also want to set the following environment variables on your shell's
+.zshrc or .bash_profile:
+
+::
+
+   export PKG_CONFIG_PATH="/usr/local/opt/openssl@1.1/lib/pkgconfig"
```
Akaricchi: Does it fail to build without this, or just doesn't pick up the openssl dependency? Note that the openssl dependency is *very* optional: currently it's only used for the SHA256 function, which is used by the shader cache. The shader cache is only used by the GLES backends right now (technically, anything that requires the shader transpiler).

But Taisei also has a built-in (possibly slower) SHA256 implementation, used as fallback if openssl is not available. This fallback can also be forced with `-Duse_libcrypto=false`. It really only makes sense to link against openssl for Linux distribution packages, where it's a core component of the system that's pretty much guaranteed to be present.

doc/DEVELOPMENT.rst
```diff
@@ -0,0 +1,266 @@
...
+tools:
+
+::
+
+    brew install meson cmake pkg-config docutils imagemagick pygments
+
+
+The following dependencies are technically optional, and can be pulled in at
+build-time, but you're better off installing them yourself to reduce compile
+times:
```
Akaricchi: It's probably worth mentioning that it's possible to force internal dependencies with `--wrapmode=forcefallback` when making the build directory. This is used when building the official releases for Windows, macOS, and Emscripten, for consistency and ease of dependency management on my end.

doc/DEVELOPMENT.rst
```diff
@@ -0,0 +1,266 @@
...
+
+Taisei's GFX library is made up of a collection of sprites, shaders, and a few
+3D models. The 3D models are almost exclusively used for background scenery
+(and a few other places, like the HUD), while the sprites are used in everything
+from UI elements, character portraits, to the
+danmaku bullets themselves.
+
+To modify the 3D models, you'll need `Blender <https://blender.org>`__, which is
+free and open source. Look for tutorials on YouTube for how to use it. The
+models themselves are located in ``resources/00-taisei.pkgdir/models``.
```
Akaricchi: Definitely worth documenting the export options for `.obj` models, since they are different from Blender's defaults.

export options

doc/DEVELOPMENT.rst
StarWitch: Hmm, okay. It does _work_ without the dependency, but I figured if anyone was building it, it'd be nice to have.

But yeah, it does introduce some overhead. I can get rid of it.

doc/DEVELOPMENT.rst
```diff
@@ -229,6 +231,22 @@ To modify the 3D models, you'll need `Blender <https://blender.org>`__, which is
...
+
+    Include
+        Objects as OBJ Objects: ENABLED
+
+    Transform
+        Forward: Y Forward
+        Up: -Z Up
+
+    Geometry
+        Write Materials: DISABLED
```
Akaricchi: Triangulation is important, Taisei will bail on quad faces. I'm pretty sure it's disabled by default, at least on my system.

README.rst
```diff
@@ -6,122 +6,86 @@ Taisei
...
-shoot-em-up games set in an isolated world full of Japanese folklore.
+Taisei Project is an original open source fan-game set in the world of Tōho
+Project. It is a top-down vertical-scrolling curtain fire shooting game (STG),
+also known as "danmaku" or "bullet hell," with an original soundtrack,
+character art, and story.
 
-Installation
-------------
+Taisei Project is written in C, with an SDL2/OpenGL renderer. It is available
+for Windows, Linux (x86_64), and macOS.
```
Akaricchi, line 15: > with an SDL2/OpenGL renderer.

This is a little misleading, because we are not using SDL for rendering. SDL *does* have an optional 2D renderer, but it's too limited for us. SDL is only used as a platform abstraction layer.

> Taisei Project is written in C, 

Perhaps it is worth mentioning that it is specifically written in C11. Unfortunately many compilers still do not support that, including Microsoft's.

> It is available for Windows, Linux (x86_64), and macOS.

Maybe something like "It works on many operating systems, including but not limited to Windows, Linux, and macOS"? Those three have the "official support", but it's known to run on some other systems as well. And it's probably worth mentioning that it can run in the browser too; that's all the rage these days apparently.

StarWitch: Done and done.

PR #192, opened 2020-02-23
New (themed) difficulty/shot mode descriptions

StarWitch:
This is just a bit of fun I think could make the Difficulty Select and shot types more in-line with the overall theme of the game ("knowledge is power"). A lot of mainline Touhou games have fun "subtitles" for their difficulty modes and shot types, like in TH13: Ten Desires, where "Lunatic Mode" has a subtitle of "Prayer For IT Data Security" (because it's "impossible", heh), and some shot types even affect the story (like in WBaWC).

I edited the .png files and regenerated the atlas. Let me know if it looks good to you, or if we want to change the font. I tried to find something "brushstroke-y" that also wasn't as intense as the main big font.

Here's the text representation of the Difficulty Modes:

Easy / Bachelor's Degree -> "For those just beginning their academic careers"
Normal / Master's Degree -> "All those years of study have finally paid off"
Hard / Ph.D -> "You must really love books"
Lunatic / Post-Doctoral -> "You either die a student, or live long enough to become a professor"


Akaricchi:
First of all I do like the idea, and I think the new subtitles and descriptions fit nicely. The "Easy" one is a bit problematic because of that pesky Y tail making it hard to read, though. 

It looks like you did not actually commit the regenerated atlas texture (should be `resources/00-taisei.pkgdir/gfx/atlas_common_ui_0.webp`), so these appear broken as they are now in-game. Also make sure it's a `.webp` and not `.png`; `ninja gen-atlas-common_ui` should generate the appropriate file.

I would've recommended editing the [high-resolution versions](https://github.com/taisei-project/taisei-rawmedia/tree/master/gfx/difficulty) of those images if I've had known you were going to touch them, but if that's a hassle to redo now it's fine.

Perhaps you could also commit the subtitle images separately into the rawmedia repo, and/or the native project files from your image manipulation software (e.g. `.psd`, `.kra`, `.xcf`, etc.). That'll help with future editing should it be ever required.

Akaricchi:
Could you also run `scripts/optimize-all-img.sh atlas/common_ui/difficulty` please?

src/plrmodes/reimu_b.c
```diff
@@ -602,7 +602,7 @@ static void reimu_dream_init(Player *plr) {
 
 PlayerMode plrmode_reimu_b = {
 	.name = "Dream Shaper",
-	.description = "Bend the boundaries of space like the ribbon in your hair.",
+	.description = "Utilize Non-Euclidean geometry you borrowed from a ’friend,’ and tie the boundaries of space-time into knots as easily as the ribbons in your hair.",
```
Akaricchi: I'm not a fan of the non-Euclidean geometry bit. Maybe it's because (and forgive me for sounding like a pedantic ass here) people tend to associate it with wacky and almost mystical properties, when in reality non-Euclidean geometry can also be quite mundane — the surface of a sphere is a non-Euclidean geometry, for example.

Also, it's probably better to spell spacetime without the dash just for consistency reasons — both seem to be valid, but one of Elly's spell cards already spells (no pun intended) it that way.

src/plrmodes/reimu_a.c
```diff
@@ -699,7 +699,7 @@ static double reimu_spirit_property(Player *plr, PlrProperty prop) {
 
 PlayerMode plrmode_reimu_a = {
 	.name = "Yōkai Buster",
-	.description = "Sometimes it is easier to let your resolve do the aiming.",
+	.description = "The tried-and-true: homing bullets and sharp needles. You don’t have the luxury of time to think about things.",
```
Akaricchi: Maybe something like this?
```suggestion
	.description = "The tried-and-true: homing amulets and sharp needles. You don’t have the luxury of time to think about the aiming.",
```

Akaricchi:
The rest of the descriptions are a nice and solid improvement in my opinion.

Akaricchi:
But you might want to rename the PR since it's not just about the difficulty levels anymore, or split the shotmodes off into another one.

StarWitch:
Changes are in! I also modified the title of this PR.

src/plrmodes/reimu_b.c
```diff
@@ -602,7 +602,7 @@ static void reimu_dream_init(Player *plr) {
 
 PlayerMode plrmode_reimu_b = {
 	.name = "Dream Shaper",
-	.description = "Bend the boundaries of space like the ribbon in your hair.",
+	.description = "Utilize techniques you borrowed from an eldritch ’friend,’ and tie the boundaries of spacetime into knots as easily as the ribbons in your hair.",
```
Akaricchi: Actually, I'm not sure we should directly mention the "borrowing" aspect here at all. Youmu's B shot doesn't mention it either, and Marisa's probably wouldn't too (when/if it's restyled after Patchy's elemental magic). My hope is that the description and the visuals of the shots are just about enough for an observant player to pick up on the idea.

If anything, we could imply that the protags obtained the B-shot powers by studying their "superiors" rather than "borrowing" them, since it lines up better with the "science" theme.

StarWitch: Changed!

Issue #193, opened 2020-02-24
[feature] "Trainer Mode" for new players


StarWitch:
I was pitching this on Discord last night.

STGs are incredibly skill-heavy games, although everyone plays them for their own individual sense of accomplishment. ZUN even this back in the [Author's Notes for PCB](https://en.touhouwiki.net/wiki/Perfect_Cherry_Blossom/Afterword): some people are content with a 1cc Normal, some people want to get a full No-Miss No-Bomb EXTRA, and some just want to watch replays on YouTube while sipping tea.

One thing that could help people get better at STG games in general would be if Taisei had a "Trainer Mode."

Basic Ideas:
1. Allow players to toggle invulnerability, and track how many times they get hit (minus successful deathbombs) - maybe even tell them where in the stage they got hit (which Spellcard, etc, so they can go practice that part in particular).
2. Allow extensions of lives/bombs, and track by how many (and where) they go over.

Advanced Ideas (probably too much work for a first iteration):
1. Replays that you can slow down/rewind to see where you messed up, to let people figure out how they could position themselves better.
2. Toggle "lines of action" where lasers and bullets are going (in Replays only) so they can visually see patterns and figure out better positioning.

But why? Well, advanced users can already download the source code and turn on `-Ddeveloper=true` and press `Q`, and there's already internal flags for setting if "cheats" were used in Replays. But, it's pretty basic (doesn't show you *how* you failed) and all-or-nothing, and you either have God Mode with unlimited everything, or nothing at all. 

There's that idea of "git gud," but even "hardcore" speedrunning games like Celeste have this sort of functionality now. Letting beginners (or even advanced players looking to get the flow of a "full" stage) toggle these features individually would make things more accessible while not taking away the challenge or allure of doing an "unassisted" 1cc.

PR #194, opened 2020-02-27
Trainer Mode 

StarWitch:
For #193 .

Right now, the following is implemented:
* **Extra Lives (when needed)** - you can still die, but only lives that you lose when you go below 0 total lives will count as "extras". If you end up collecting extra lives, those get used first.
* **Extra Spellcards (when needed)** - same as above. Stored used first, extras given/tracked.
* **Invulnerability (with Hit Tracking)** - lets you turn on God Mode (like ``iddqd``) but will track whenever you get hit by a bullet.
* **No Powerdown When Hit** - if you die, you don't lose any of your Power gauge. 
* **HUD Stats & Tracking** - toggle whether or not you want to see stats in the HUD while playing the game. Some people can find them distracting. There is also a "per stage" and "per play-through" tracking extra lives/extra bombs/misses while invulnerable. That way you can see which stages trip you up the most, and how, and also overall stats for your entire run.

I broke it out into its own module so that additional functionality can be added easily enough down the line. This is a minimum viable implementation right now. 

The HUD display isn't the most polished but it can be changed easily. I put it in there just to see what it'd look like. I'm also hoping to eventually implement the ability to log when/where you end up dying, getting hit, or needing an extra bomb (which Spellcard, etc) and track stats on a menu entry somewhere.

I don't think adding a "Slowdown" feature is worth it at the moment, mostly because of how much work is being done on Coroutines and how screwy global timers are. We'd also have to slow down audio clips that play, which as far as I know isn't a thing that's implemented right now. And at the end of the day, either someone's using it to get better at the game (or STGs in general), or they're using it so they can see the whole game at their leisure, and in either case Slowdown doesn't really help them do that.


src/trainer.c
```diff
@@ -0,0 +1,92 @@
...
+    tnr->trainer_hits_stage = 0;
+
+    tnr->trainer_lives_total = 0;
+    tnr->trainer_bombs_total = 0;
+    tnr->trainer_hits_total = 0;
+}
+
+
+bool trainer_enabled(void) {
+    if (config_get_int(CONFIG_TRAINER_MODE) && config_get_int(CONFIG_TRAINER_STATS)) {
```
Akaricchi: I'm a bit confused about the `CONFIG_TRAINER_STATS` setting. What is it for exactly?

src/trainer.h
```diff
@@ -0,0 +1,37 @@
...
+
+struct Trainer {
+	int trainer_lives_stage;
+	int trainer_bombs_stage;
+	int trainer_hits_stage;
+
+	int trainer_lives_total;
+	int trainer_bombs_total;
+	int trainer_hits_total;
+};
```
Akaricchi: I would recommend something like that instead:

```suggestion
struct Trainer {
	struct {
		int lives;
		int bombs;
		int hits;
	} total, stage;
};
```

Keeping total and stage stats in separate sub-structs like this allows easily resetting them separately with a memset call:
```c
memset(&trailer.stage, 0, sizeof(trainer.stage));
```

The `trainer_` prefix on all fields is also unnecessary.

src/global.h
```diff
@@ -135,6 +136,8 @@ typedef struct {
 
 	StageInfo *stage;
 
+	Trainer tnr;
```
Akaricchi: The global object is bad enough; please rename this field to `trainer` for clarity.

src/trainer.c
```diff
@@ -0,0 +1,92 @@
...
+        if (config_get_int(CONFIG_TRAINER_LIVES) || config_get_int(CONFIG_TRAINER_BOMBS) || config_get_int(CONFIG_TRAINER_INVULN)) {
+            return 1;
```
Akaricchi: Please use the `true` and `false` macros for booleans.

src/player.c
```diff
@@ -923,6 +938,12 @@ void player_death(Player *plr) {
 		return;
 	}
 
+	if (trainer_invulnerable_enabled()) {
+		trainer_append_hit_event(&global.tnr);
+		play_sound("shot3");
```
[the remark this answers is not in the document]
StarWitch: I was thinking of more of a "ping" sound. I'll try to find one that's more distinct but also low-key.

src/stagedraw.c
```diff
@@ -1255,6 +1268,14 @@ static void stage_draw_hud_text(struct labels_s* labels) {
 	draw_label("Value:",       labels->y.value,   labels, &stagedraw.hud_text.color.label_value);
 	draw_label("Volts:",       labels->y.voltage, labels, &stagedraw.hud_text.color.label_voltage);
 	draw_label("Graze:",       labels->y.graze,   labels, &stagedraw.hud_text.color.label_graze);
+
+	if (trainer_enabled()) {
+		draw_label("Trainer Mode", labels->y.trainer, labels, &stagedraw.hud_text.color.label_trainer);
+		draw_label("Lives:", labels->y.trainer_lives, labels, &stagedraw.hud_text.color.label_trainer_lives);
+		draw_label("Spells:", labels->y.trainer_bombs, labels, &stagedraw.hud_text.color.label_trainer_bombs);
+		draw_label("Hits:", labels->y.trainer_hits, 0, &stagedraw.hud_text.color.label_trainer_hits);
```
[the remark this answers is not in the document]
StarWitch: Actually a typo here. Fixed anyways!

src/trainer.c
StarWitch: It determines whether or not the stats are shown in real-time on the HUD, or kept strictly to the Stage Clear screen - I'm sure some people might find them distracting. However, its usage here wasn't quite correct, but I've fixed it.

src/player.c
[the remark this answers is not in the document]
StarWitch: OK, I added ``tick.opus``. It sounds distinct but isn't annoying, and is also CC-0/Public Domain. 

https://freesound.org/people/NenadSimic/sounds/268108/

src/trainer.c
```diff
@@ -14,78 +14,92 @@
...
-
 bool trainer_enabled(void) {
-	if (config_get_int(CONFIG_TRAINER_MODE) && config_get_int(CONFIG_TRAINER_STATS)) {
-		if (config_get_int(CONFIG_TRAINER_LIVES) || config_get_int(CONFIG_TRAINER_BOMBS) || config_get_int(CONFIG_TRAINER_INVULN)) {
-			return 1;
-		}
+	if (config_get_int(CONFIG_TRAINER_MODE)) {
+		return true;
+	}
+	return false;
```
Akaricchi: Actually, why not just return the expression inside the `if` from those functions? Same with the new `player_can_bomb`, etc.

Another style nitpick that I missed: we don't put spaces before the opening parenthesis after `if`/`for`/`while`, sorry.

src/player.c
```diff
@@ -850,19 +862,21 @@ void player_realdeath(Player *plr) {
...
 	if (!trainer_no_powerdown_enabled()) {
-		// if a player gets hit with trainer-no_pwrdn enabled...
+		// if "no powerdown" trainer is DISABLED, then powerdown as usual
 		int drop = max(2, (total_power * 0.15) / POWER_VALUE);
 		spawn_items(plr->deathpos, ITEM_POWER, drop);
 		player_set_power(plr, total_power * 0.7);
 		plr->voltage *= 0.9;
-		log_debug("Trainer Mode - No Powerdown");
+	} else {
+		// log when "no powerdown" trainer used
```
Akaricchi: The code is self-explanatory so I don't think this comment is needed.

StarWitch:
Yeah, replays were going to be the next thing I worked on with respect to this. As I said, this isn't a "complete" implementation, it's a basic one to see how it looks and fits in with the game.

I think the idea of having to set which options you want before you start playing is a decent one, as it's also something I thought of, not only for replays but just with the spirit of the mode itself in general.

----

As for being able to see the good endings and all that... I think Trainer Mode should still allow it. But let me explain first.

As I said in the original Issue ticket, a lot of "hardcore" games are letting players do that these days. Celeste is my big inspiration for this in Taisei: it's an incredibly intense and competitive masocore platformer, where you're expected to die hundreds if not thousands of times in a casual playthrough, but you can still get to the end of the game and unlock everything with their Assist Mode turned on.

It encourages and incentivizes players to beat the game with Assist Mode turned off, though. It adds a badge to your completion screens saying that Assist Mode was used, which goes away when you beat each Chapter/Stage with it turned off. There's also messages throughout the game telling players that it's supposed to be challenging, and that the challenge is part of the genre (masocore). We could do the same thing with "Trainer Mode" enabled on the HUD as a mandatory feature (with only stats being toggleable), as well as having it marked as "Cheats Used" in replays, which is already a thing due to ``iddqd``.

Also, from a personal perspective, doing stage/spell practice is a different feel from playing through the game from start to finish. I play differently due to the flow of the stages and the feeling of momentum in a good run. It's hard to achieve that "zen" state when the start and stops are so sudden.

Anyways, that's just my opinion.

Akaricchi:
Alright, I see where you're coming from. I do agree that it should cheat-mark the replays and display the HUD message. We can also track stage & game clears with trainer mode/cheats separately in the progress module, perhaps for a future stats screen. In that case I'd be ok with it.

Akaricchi:
This system could be further extended later with options that not just make the game easier, but also harder, sillier, or just different somehow (various challenge modes, deadly pickups, permanent Mystia blindness, UFOs from th12, you name it). I actually thought of something like that before.

StarWitch:
A challenge mode would be a lot of fun. A few of my friends actually complain that Touhou-themed STGs are "too easy" even on Lunatic so adding additional challenge options would be a cool idea just to show off.

StarWitch:
I redid a bit of this to make it fit in better with the overall structure of the code. It also now works with replays, displays what replays had Trainer Mode enabled in the replay select menu, as well as displaying it in the character select. You also can't select it from in-game to prevent people flipping the settings during a run and potentially messing with replay stats - either you start a run with it, or you don't. It's not a particularly long game so I don't think this is too big of a deal.

StarWitch:
I rebased this branch to get rid of the merge conflicts. Hopefully everything looks as it should now.

StarWitch:
Closing this PR for now since we're going to have to think about rethinking the way game sessions are handled before v1.4, to make things easier to maintain.

PR #200, opened 2020-03-17
Implement Robin Hood hashing

Akaricchi:
This replaces the previous separate chaining implementation. This approach is more memory-efficient, cache-friendly, and puts much less stress on the memory allocator.


Akaricchi:
yolo

PR #204, opened 2020-03-21
Video resolution improvements/high-DPI quirks

StarWitch:
Some changes to hopefully make Taisei look better on high-DPI displays. 

(Still needs some work, but I'm putting it up now for posterity.)

### DIFF
--- a/src/video.c
+++ b/src/video.c
@@ -126,9 +126,13 @@ static int video_compare_modes(const void *a, const void *b) {
 	return va->width * va->height - vb->width * vb->height;
 }
 
+static bool video_is_highdpi(void) {
+	return SDL_GetWindowFlags(video.window) & SDL_WINDOW_ALLOW_HIGHDPI;
+}
+
 void video_get_viewport_size(float *width, float *height) {
-	float w = video.current.width;
-	float h = video.current.height;
+	int w, h;
+	SDL_GL_GetDrawableSize(video.window, &w, &h);
 	float r = w / h;
 	if(r > VIDEO_ASPECT_RATIO) {
 		w = h * VIDEO_ASPECT_RATIO;
@@ -142,8 +146,13 @@ void video_get_viewport_size(float *width, float *height) {
 
 void video_get_viewport(FloatRect *vp) {
 	video_get_viewport_size(&vp->w, &vp->h);
-	vp->x = (int)((video.current.width  - vp->w) / 2);
-	vp->y = (int)((video.current.height - vp->h) / 2);
+	if(video_is_highdpi()) {
+		vp->x = (int)(video.current.width  - (vp->w / 2));
+		vp->y = (int)(video.current.height - (vp->h / 2));
+	} else {
+		vp->x = (int)((video.current.width  - vp->w) / 2);
+		vp->y = (int)((video.current.height - vp->h) / 2);
+	}
 }
 
 static void video_set_viewport(void) {
@@ -248,7 +257,9 @@ static void video_new_window(uint display, uint w, uint h, bool fs, bool resizab
 		flags |= SDL_WINDOW_RESIZABLE;
 	}
 
-	flags |= SDL_WINDOW_ALLOW_HIGHDPI;
+	if(video.backend == VIDEO_BACKEND_COCOA) {
+		flags |= SDL_WINDOW_ALLOW_HIGHDPI;
+	}
 
 	video_new_window_internal(display, w, h, flags, false);
 	display = video_current_display();
@@ -683,6 +694,8 @@ void video_init(void) {
 	} else if(!strcmp(driver, "Switch")) {
 		video.backend = VIDEO_BACKEND_SWITCH;
 		video_query_capability = video_query_capability_alwaysfullscreen;
+	} else if(!strcmp(driver, "Cocoa")) {
+		video.backend = VIDEO_BACKEND_COCOA;
 	} else {
 		video.backend = VIDEO_BACKEND_OTHER;
 	}
--- a/src/video.c
+++ b/src/video.c
@@ -19,8 +19,8 @@
 #include "video_postprocess.h"
 
 static struct {
-	VideoMode fs_modes;
-	VideoMode win_modes;
+	VideoMode *fs_modes;
+	VideoMode *win_modes;
 	SDL_Window *window;
 	uint win_mcount;
 	uint fs_mcount;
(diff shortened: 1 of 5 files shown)


src/video.c
```diff
@@ -112,11 +126,14 @@ static int video_compare_modes(const void *a, const void *b) {
...
+static bool video_is_highdpi(void) {
+	return SDL_GetWindowFlags(video.window) & SDL_WINDOW_ALLOW_HIGHDPI;
+}
+
 void video_get_viewport_size(float *width, float *height) {
-	float w = video.current.width;
-	float h = video.current.height;
+	int w, h;
+	SDL_GL_GetDrawableSize(video.window, &w, &h);
 	float r = w / h;
```
Akaricchi: You changed the types of `w` and `h` to `int`, so `w / h` is now integer division, which is wrong. You have to cast one of the operands to `float` here.
```suggestion
	float r = (float)w / h;
```

src/video.c
```diff
@@ -129,8 +146,13 @@ void video_get_viewport_size(float *width, float *height) {
...
 	video_get_viewport_size(&vp->w, &vp->h);
-	vp->x = (int)((video.current.width  - vp->w) / 2);
-	vp->y = (int)((video.current.height - vp->h) / 2);
+	if(video_is_highdpi()) {
+		vp->x = (int)(video.current.width  - (vp->w / 2));
+		vp->y = (int)(video.current.height - (vp->h / 2));
+	} else {
+		vp->x = (int)((video.current.width  - vp->w) / 2);
+		vp->y = (int)((video.current.height - vp->h) / 2);
+	}
```
Akaricchi: I have no idea what is going on in the true branch here, but I'm pretty sure it's wrong. First of all you can't use `video.current` here anymore, since the viewport size is calculated from `SDL_GL_GetDrawableSize` now, which will have a different scaling factor in high DPI mode.

The re-arranged formula in the true branch doesn't make sense either. This is supposed to implement letterboxing when the intended viewport's aspect ratio (constant 4:3 in our case) doesn't match that of the actual window. An offset is calculated such as to position the viewport in the center of the screen. This is done by subtracting the center of the viewport from the center of the window (and to find the center, you just divide the dimensions by two, hence the factored out `/2`).

If this actually looks ok on your system, I suspect it's only by sheer accident, and a different scaling factor will probably break it.

src/video.c
```diff
@@ -235,6 +257,10 @@ static void video_new_window(uint display, uint w, uint h, bool fs, bool resizab
 		flags |= SDL_WINDOW_RESIZABLE;
 	}
 
+	if(video.backend == VIDEO_BACKEND_COCOA) {
+		flags |= SDL_WINDOW_ALLOW_HIGHDPI;
+	}
```
Akaricchi: I think we should support DPI scaling on any platform that SDL can handle. Let's remove this condition and see what breaks.

src/video.c
```diff
@@ -90,20 +95,29 @@ static VideoCapabilityState video_query_capability_webcanvas(VideoCapability cap
...
+	mode_array->mode[mode_array->mcount-1].height = height;
+}
+
+static void video_add_mode(int width, int height, bool fullscreen) {
+	if(fullscreen) {
+		video_add_mode_handler(&video.fs_modes, width, height);
+	} else {
+		video_add_mode_handler(&video.win_modes, width, height);
+	}
 }
```
Akaricchi: I'd rather have separate `video_{add,get}_{fullscreen,windowed}_mode` functions than branch on a boolean like that. It's not clear what `video_add_mode(800, 600, true)` means without context.

StarWitch: I was worried about duplicating code, but if separating them makes it easier, I'm totally fine with that.

src/video.c
```diff
@@ -129,8 +146,18 @@ void video_get_viewport_size(float *width, float *height) {
...
 void video_get_viewport(FloatRect *vp) {
 	video_get_viewport_size(&vp->w, &vp->h);
-	vp->x = (int)((video.current.width  - vp->w) / 2);
-	vp->y = (int)((video.current.height - vp->h) / 2);
+	if(video_highdpi_mode()) {
+		vp->x = (int)(video.current.width  - vp->w);
+		vp->y = (int)(video.current.height - vp->h);
+		vp->w = (int)(vp->w * 2);
+		vp->h = (int)(vp->h * 2);
+	} else {
```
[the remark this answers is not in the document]
Akaricchi: Now, another thing: **if I'm right about all of the above**, then maybe we actually do **not** want to use `SDL_WINDOW_ALLOW_HIGHDPI`, ever. It would make sense to simply tell the operating system to not automatically scale the window (in macOS's case, via the `Info.plist` entry as you have already done), and then have the SDL screen-space functions like `SDL_GetWindowSize` and `SDL_CreateWindow` operate in real pixel space, without magical implicit scaling. A benefit of this is that the displayed resolution in the options menu will actually be correct, and this should work on more platforms (SDL doesn't seem to have magic DPI scaling for Windows and X11 at least).

`SDL_GL_GetDrawableSize` would also coincide with `SDL_GetWindowSize` once again, although I think it's still more technically correct to express the viewport calculation in terms of the framebuffer size explicitly.

A downside of this approach is that we'll need to do a bit of extra work to make sure that the default window is not tiny. This could be done by pruning tiny windowed modes based on `SDL_GetDisplayDPI`, using about 100dpi as the baseline (roughly corresponds to a common 1920x1080 22" monitor). Or we might as well just default to fullscreen. Either way, that's a somewhat separate issue and can be dealt with later.

WebGL is still a mystery, whether `SDL_WINDOW_ALLOW_HIGHDPI` is used or not.

src/video.c
```diff
@@ -668,6 +711,8 @@ void video_init(void) {
 	} else if(!strcmp(driver, "Switch")) {
 		video.backend = VIDEO_BACKEND_SWITCH;
 		video_query_capability = video_query_capability_alwaysfullscreen;
+	} else if(!strcmp(driver, "cocoa")) {
+		video.backend = VIDEO_BACKEND_COCOA;
```
Akaricchi: Looks like this is not used anymore?

StarWitch: Removed!

src/video.h
```diff
@@ -45,6 +45,7 @@ typedef enum VideoBackend {
 	VIDEO_BACKEND_KMSDRM,
 	VIDEO_BACKEND_RPI,
 	VIDEO_BACKEND_SWITCH,
+	VIDEO_BACKEND_COCOA,
```
Akaricchi: forgot this one

StarWitch: Done.

PR #207, opened 2020-04-03
Add generic type-safe facility for dynamic arrays

Akaricchi:
Replace most ad-hoc opencoded dynamic arrays across the codebase

### DIFF
--- a/src/ending.c
+++ b/src/ending.c
@@ -12,16 +12,16 @@
 #include "global.h"
 #include "video.h"
 #include "progress.h"
+#include "dynarray.h"
 
 typedef struct EndingEntry {
-	char *msg;
+	const char *msg;
 	Sprite *sprite;
 	int time;
 } EndingEntry;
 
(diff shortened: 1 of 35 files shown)


src/ending.c
```diff
@@ -330,29 +326,28 @@ void good_ending_reimu(Ending *e) {
...
+
 	if(global.plr.continues_used) {
 		global.plr.mode->character->ending.bad(e);
 	} else {
 		global.plr.mode->character->ending.good(e);
 		add_ending_entry(e, 400, "-UNDER CONSTRUCTION-\nExtra Stage coming soon! Please wait warmly for it!", NULL);
 	}
 
-	add_ending_entry(e, 400, "", NULL); // this is important
+	add_ending_entry(e, 400, "", NULL);  // this is important   <-- TODO explain *why* this is important?
```
lukas-weber, line 338: I think this is like the sentinel element, but i’m not sure.

PR #211, opened 2020-04-12
Turn on SDL_WINDOW_ALLOW_HIGHDPI

holycleugh:
Hi, I needed this flag to remove the blurriness on my setup (Wayland, 3840x2160 monitor with scale factor 2). 

I'm not sure if you want this flag on conditionally or somewhere else or as a build parameter or something, so I can move this assignment or change the logic - just let me know.

In the screenshot below, the left build is with the patch, and the right one is master.

20200412_01h19m03s_grim

### DIFF
--- a/src/video.c
+++ b/src/video.c
@@ -257,6 +257,8 @@ static void video_new_window_internal(uint display, uint w, uint h, uint32_t fla
 static void video_new_window(uint display, uint w, uint h, bool fs, bool resizable) {
 	uint32_t flags = 0;
 
+	flags |= SDL_WINDOW_ALLOW_HIGHDPI;
+
 	if(fs) {
 		flags |= get_fullscreen_flag();
 	} else if(resizable && video.backend != VIDEO_BACKEND_EMSCRIPTEN) {


Akaricchi:
Thanks for reporting. However, some more changes will be required to reconcile this with some of the issues mentioned in the discussion of #204.

We should indeed be using this flag after all, but that means we have to divorce the framebuffer resolution from the abstract screen-space coordinates as used by SDL's window functions, such as `SDL_CreateWindow` and `SDL_SetWindowSize`. It is currently assumed that these correspond to the raw screen resolution in pixels, but that assumption is violated when `SDL_WINDOW_ALLOW_HIGHDPI` and a scaling factor is involved (at least on some platforms). This leads to some usability problems, like a subtly broken resolution selection in the options menu.

For Windows and macOS we currently deal with this problem by explicitly disabling OS DPI scaling via the manifest/`Info.plist` respectively, and not using `SDL_WINDOW_ALLOW_HIGHDPI`. This way we get to work with raw pixel sizes as before, without OS-side upscaling, regardless of the monitor DPI/scaling factor. This seems to work for X11 as well (there seems to be no upscaling to begin with, from what I can tell). Wayland is a different story - apparently the flag is required to disable the framebuffer upscaling (and it's probably compositor-dependent as well).

cc @StarWitch

Akaricchi:
Please tell me what Wayland compositor you're using; I'd like to reproduce this.

holycleugh:
@Akaricchi Understood - feel free to close this PR if it'll need a much larger change to the game code. Indeed the resolution setting in the game shows 1920x1080.

I'm using wlroots/Sway. Can get you more information if you need it.

```
~ swaymsg -t get_outputs
Output HDMI-A-1 'Goldstar Company Ltd LG Ultra HD 0x0000EEA0' (focused)
  Current mode: 3840x2160 @ 60.000000 Hz
  Position: 0,0
  Scale factor: 2.000000
  Scale filter: nearest
  Subpixel hinting: unknown
  Transform: normal
  Workspace: 2venturus
  Max render time: off
```

PR #213, opened 2020-04-16
Improve high-DPI handling

Akaricchi:
`SDL_WINDOW_ALLOW_HIGHDPI` is now used. On platforms where SDL supports this flag, the window size units are interpreted as abstract scaled points rather than raw pixels, and may not match the underlying framebuffer resolution if the display scaling factor is not 1. The scaling factor and the effective framebuffer resolution are now displayed in the options menu to reflect this fact.

In addition, the hidden `fullscreen_desktop_mode` setting has been removed and the behavior is now always as if the value was `1` (the default). Exclusive fullscreen with modesetting is not very useful, and does not work well on some platforms (X11 in particular). That code is not worth maintaining together with the frankly ridiculous complexity of dancing around with the DPI scaling shenanigans of various operating systems.

On platforms that do not support `SDL_WINDOW_ALLOW_HIGHDPI`, the window size is assumed to correspond to raw pixels, and an effort is made to disable the operating system's DPI upscaling mechanism, when possible.

This commit also fixes the Windows manifest, broken by dc57f78d89f654344ab2314a05f459fde0b76bda thanks to microshaft's amazing documentation.

Closes #211

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -89,7 +89,6 @@
 	\
 	CONFIGDEF_STRING    (PLAYERNAME,                "playername",                           "Player") \
 	CONFIGDEF_INT       (FULLSCREEN,                "fullscreen",                           0) \
-	CONFIGDEF_INT       (FULLSCREEN_DESKTOP,        "fullscreen_desktop_mode",              1) \
 	CONFIGDEF_INT       (VID_WIDTH,                 "vid_width",                            RESX) \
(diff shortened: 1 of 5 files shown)


Akaricchi:
@holycleugh I've tested this under Sway with no issues, but please confirm if this works for you.

@StarWitch and @p-sam, please test this for macOS and Switch regressions, respectively.

holycleugh:
Sorry for the delay. Looks great on my computer, thank you!

PR #219, opened 2020-04-26
Player Statistics

StarWitch:
Forked off of #194, as per our discussion in Discord.

Not much has changed except for the fact that now I'm using the REPLAY_GFLAG* typedef to save booleans rather than writing a bunch of booleans to the replay file one at a time.

This way we can also have additional game modes, assist mode options, "comedy options" etc. in their own separate framework without being tied together with this.

Has some rough edges but I'm mostly looking for a yay/nay before I clean it up.

### DIFF
--- a/src/menu/common.c
+++ b/src/menu/common.c
@@ -91,6 +91,7 @@ static void reset_game(StartGameContext *ctx) {
 	replay_destroy(&global.replay);
 	replay_init(&global.replay);
 	player_init(&global.plr);
+	stats_init(&global.plr.stats);
 	global.plr.mode = plrmode_find(
 		progress.game_settings.character,
 		progress.game_settings.shotmode
--- a/src/player.c
+++ b/src/player.c
@@ -16,6 +16,7 @@
 #include "stage.h"
 #include "stagetext.h"
 #include "stagedraw.h"
+#include "stats.h"
 #include "entity.h"
 #include "util/glm.h"
 
@@ -590,7 +591,7 @@ DEFINE_TASK(player_logic) {
 			plr->point_item_value = PLR_START_PIV;
 			plr->life_fragments = 0;
 			plr->bomb_fragments = 0;
-			plr->continues_used += 1;
+			stats_append_continue(&plr->stats);
 			player_set_power(plr, 0);
 			stage_clear_hazards(CLEAR_HAZARDS_ALL);
 			spawn_items(plr->deathpos, ITEM_POWER, (int)ceil(PLR_MAX_POWER/(double)POWER_VALUE));
@@ -638,6 +639,19 @@ DEFINE_TASK(player_logic) {
 void player_logic(Player* plr) {
 }
 
+static bool player_can_bomb(Player *plr) {
+	return (
+			!player_is_bomb_active(plr)
+			&& (
+				plr->bombs > 0 ||
+				plr->iddqd
+			)
+			&& global.frames >= plr->respawntime
+	);
+
+}
+
+
 static bool player_bomb(Player *plr) {
 	if(global.boss && global.boss->current && global.boss->current->type == AT_ExtraSpell)
 		return false;
@@ -648,7 +662,8 @@ static bool player_bomb(Player *plr) {
 		return false;
 	}
 
-	if(!player_is_bomb_active(plr) && (plr->bombs > 0 || plr->iddqd) && global.frames >= plr->respawntime) {
+	if(player_can_bomb(plr)) {
+		stats_append_bomb(&plr->stats);
 		player_fail_spell(plr);
 		// player_cancel_powersurge(plr);
 		// stage_clear_hazards(CLEAR_HAZARDS_ALL);
@@ -885,6 +900,7 @@ void player_realdeath(Player *plr) {
 	plr->bomb_fragments = 0;
 	plr->voltage *= 0.9;
 	plr->lives--;
+	stats_append_life(&plr->stats);
 }
 
 static void player_death_effect_draw_overlay(Projectile *p, int t, ProjDrawRuleArgs args) {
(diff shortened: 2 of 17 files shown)


src/replay.h
```diff
@@ -130,6 +133,15 @@ typedef struct ReplayStage {
 	uint64_t plr_points_final;
 	/* END REPLAY_STRUCT_VERSION_TS102000_REV3 and above */
 
+	/* BEGIN REPLAY_STRUCT_VERSION_TS104000_REV0 and above */
+	uint8_t plr_stats_enabled;
```
Akaricchi, line 137: This one is not used I think?

src/stagedraw.c
```diff
@@ -1826,7 +1826,7 @@ void stage_display_clear_screen(const StageClearBonus *bonus) {
 
 	stagetext_add(
 		"Press Fire to continue",
-		VIEWPORT_W/2 + VIEWPORT_H*0.7*I,
+		VIEWPORT_W/2 + VIEWPORT_H*0.8*I,
```
Akaricchi, line 1829: Unrelated change?

src/stagetext.c
```diff
@@ -187,16 +187,6 @@ void stagetext_table_add_separator(StageTextTable *tbl) {
...
-void stagetext_table_test(void) {
-	StageTextTable tbl;
-	stagetext_begin_table(&tbl, "Test", RGB(1, 1, 1), RGB(1, 1, 1), VIEWPORT_W/2, 60, 300, 30, 60);
-	stagetext_table_add(&tbl, "foo", "bar");
-	stagetext_table_add(&tbl, "qwerty", "asdfg");
-	stagetext_table_add(&tbl, "top", "kek");
-	stagetext_table_add_separator(&tbl);
-	stagetext_table_add_numeric(&tbl, "Total Score", 9000000);
-	stagetext_end_table(&tbl);
-}
```
Akaricchi, line 199: This is also unrelated, but it's dead code so I don't mind it gone; I guess it's fine.

src/stats.c
```diff
@@ -0,0 +1,48 @@
...
+ * This software is licensed under the terms of the MIT License.
+ * See COPYING for further information.
+ * ---
+ * Copyright (c) 2011-2019, Lukas Weber <laochailan@web.de>.
+ * Copyright (c) 2012-2019, Andrei Alexeyev <akari@taisei-project.org>.
+ */
+
+#include "taisei.h"
+
+#include "config.h"
```
Akaricchi, line 11: Not needed

src/stats.c
```diff
@@ -0,0 +1,48 @@
...
+
+#include "stats.h"
+
+#include "global.h"
+#include "stage.h"
```
Akaricchi, line 16: Probably also not needed. For the log stuff, just include `log.h`.

src/stats.h
```diff
@@ -0,0 +1,32 @@
...
+#ifndef IGUARD_stats_h
+#define IGUARD_stats_h
+
+#include "taisei.h"
+
+typedef struct Stats Stats;
+
+struct Stats {
+	struct {
+		int lives;
```
Akaricchi, line 18: I think this is a little misleading, because it can be interpreted as either "lives used" or "total lives gained". I would rename it either `deaths` or `lives_used`; perhaps the same applies to bombs.

src/player.c
```diff
@@ -885,6 +900,7 @@ void player_realdeath(Player *plr) {
 	plr->bomb_fragments = 0;
 	plr->voltage *= 0.9;
 	plr->lives--;
+	stats_append_life(&plr->stats);
```
Akaricchi, line 903: Case in point: I was pretty confused about this code fragment, appending a "life" when a "death" has occurred, until I realized what the correct interpretation is.

src/stats.h
```diff
@@ -0,0 +1,32 @@
...
+		int bombs;
+		int continues;
+	} total, stage;
+};
+
+void stats_init(Stats *stats);
+
+void stats_append_life(Stats *stats);
+void stats_append_bomb(Stats *stats);
+void stats_append_continue(Stats *stats);
```
Akaricchi, line 28: Also, I would suggest instead of "append" to use the word "track", as in `stats_track_<event>`. `stats_track_life_lost`, `stats_track_bomb_used`, reads more naturally IMO.

src/menu/common.c
```diff
@@ -91,6 +91,7 @@ static void reset_game(StartGameContext *ctx) {
 	replay_destroy(&global.replay);
 	replay_init(&global.replay);
 	player_init(&global.plr);
+	stats_init(&global.plr.stats);
```
Akaricchi, line 94: I think you also need this call in `main_singlestg_begin_game` (`main.c`). That is where game initialization is done when starting a stage from the command line (`taisei --play --sid <stageID>`).

(Ideally we should really improve our terrible game-session abstraction (or lack thereof), but that's out of scope for this PR)

src/stats.c
```diff
@@ -0,0 +1,48 @@
...
+	stats->total.continues++;
+	stats->stage.continues++;
+	log_debug("bombs used (total): %d", stats->total.continues);
+	log_debug("bombs used (stage): %d", stats->stage.continues);
+}
+
+void stats_stage_reset(Stats *stats) {
+	stats->stage.lives = 0;
+	stats->stage.bombs = 0;
+	stats->stage.continues = 0;
```
Akaricchi, line 46:
```suggestion
	memset(&stats->stage, 0, sizeof(stats->stage));
```

src/stats.c
Akaricchi: Clarification: this comment only applies to the `#include "config.h"` line; `taisei.h` must be included by every file.

StarWitch:
Current round of review done! Some of that was just me being tired and sloppy. Thanks for catching it.

PR #220, opened 2020-04-26
Title Cards

StarWitch:
Still very much a WIP, and very rough. This gives the boss characters "title cards" which pop in when queued in the dialogue, and then fade out after a preset time, like mainline Touhou games.

Since this is a highly-visual sort of change, I'm looking for feedback in terms of style, position, animation, etc. and anything else we want this bit of flair to do.

### DIFF
--- a/src/dialog.c
+++ b/src/dialog.c
@@ -335,12 +335,10 @@ void dialog_draw(Dialog *dialog) {
 
 	r_mat_mv_push();
 
-	float opacity = dialog->opacity;
-
-	if(opacity < 1) {
-		r_mat_mv_translate(0, 100 * (1 - opacity), 0);
+	if(dialog->opacity < 1) {
+		r_mat_mv_translate(0, 100 * (1 - dialog->opacity), 0);
 	}
-	r_color4(0, 0, 0, 0.8 * opacity);
+	r_color4(0, 0, 0, 0.8 * dialog->opacity);
 	r_mat_mv_push();
 	r_mat_mv_translate(dialog_bg_rect.x, dialog_bg_rect.y, 0);
 	r_mat_mv_scale(dialog_bg_rect.w, dialog_bg_rect.h, 1);
@@ -360,7 +358,7 @@ void dialog_draw(Dialog *dialog) {
 
 	if(dialog->text.fading_out->opacity > 0) {
 		clr = dialog->text.fading_out->color;
-		color_mul_scalar(&clr, opacity);
+		color_mul_scalar(&clr, dialog->opacity);
 
 		text_draw_wrapped(dialog->text.fading_out->text, dialog_bg_rect.w, &(TextParams) {
 			.shader = "text_dialog",
@@ -376,7 +374,7 @@ void dialog_draw(Dialog *dialog) {
 
 	if(dialog->text.current->opacity > 0) {
 		clr = dialog->text.current->color;
-		color_mul_scalar(&clr, opacity);
+		color_mul_scalar(&clr, dialog->opacity);
 
 		text_draw_wrapped(dialog->text.current->text, dialog_bg_rect.w, &(TextParams) {
 			.shader = "text_dialog",
--- a/src/dialog.c
+++ b/src/dialog.c
@@ -107,17 +107,19 @@ void dialog_update(Dialog *d) {
 	if (d->title.active) {
 		if(d->title.timeout > 0) {
 			fapproach_asymptotic_p(&d->title.box.opacity, 1, 0.05, 1e-3);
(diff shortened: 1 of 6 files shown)


src/dialog.c
```diff
@@ -95,6 +95,24 @@ void dialog_update(Dialog *d) {
...
+			fapproach_asymptotic_p(&d->title.box.opacity, 1, 0.05, 1e-3);
+			if(d->title.box.opacity >= 0.9 ) {
+				fapproach_asymptotic_p(&d->title.box_text.opacity, 1, 0.05, 1e-3);
+			}
+			d->title.timeout--;
+		}
+		if(d->title.timeout == 0 || d->state == DIALOG_STATE_FADEOUT) {
+			fapproach_asymptotic_p(&d->title.box.opacity, 0, 0.1, 1e-3);
+			fapproach_asymptotic_p(&d->title.box_text.opacity, 0, 0.1, 1e-3);
+            d->title.box_text.opacity = 0;
```
Akaricchi, line 110: Misaligned indentation

src/dialog.c
```diff
@@ -275,8 +292,6 @@ void dialog_draw(Dialog *dialog) {
 		return;
 	}
 
-	float o = dialog->opacity;
```
Akaricchi, line 278: Not sure why this was changed. If it's for readability purposes, then consider re-introducing the variable and renaming it `opacity`. I can see how `o` may be unclear.

It is often the case in drawing code that alpha/opacity-style state variables are filtered through a function rather than consumed directly (often using one of the [easing functions](https://github.com/taisei-project/cglm/blob/ee15b56bd90941c9c27dd409df4802490645f367/include/cglm/ease.h)). While this doesn't apply here, I think it's nice to cache stuff like this into local variables by convention, to facilitate such changes.

src/dialog.c
```diff
@@ -422,3 +477,13 @@ bool dialog_is_active(Dialog *d) {
...
+
+void dialog_draw_title(Dialog *dialog, DialogActor *actor, char *name, char *title) {
+	dialog->title.name = name;
+	dialog->title.text = title;
+	dialog->title.active = true;
+	dialog->title.timeout = 360;
+
+	log_debug("Show Title: %s - %s", name, title);
+
+}
```
Akaricchi, line 489: This function does not draw anything, it only sets up state. It should be renamed `dialog_show_title` or similar.

The `draw` verb is generally reserved for functions that do rendering and do not change state. Any existing counter-example to this convention should be considered a bug.

src/dialog.h
```diff
@@ -55,6 +55,10 @@ typedef struct DialogTextBuffer {
 	float opacity;
 } DialogTextBuffer;
 
+struct title_box {
+	float opacity;
+};
```
Akaricchi, line 60: Since this is in a public header, struct definitions are generally held up to a higher standard. The name should be CamelCase and prefixed, e.g. `DialogTitleBox`, and also have a typedef with the same name.

In this specific case however, the definition is probably not even necessary: you could just define an anonymous struct inside Dialog to achieve the same effect without namespace pollution (see next comment).

src/dialog.h
```diff
@@ -64,11 +68,22 @@ typedef struct Dialog {
 		DialogTextBuffer *fading_out;
 	} text;
 
+	struct {
+		const char *name;
+		const char *text;
+		bool active;
+		int timeout;
+		struct title_box box;
+		struct title_box box_text;
```
Akaricchi, line 77:
```suggestion
		struct {
			float opacity;
		} box, box_text;
```

src/dialog.h
```diff
@@ -64,11 +68,22 @@ typedef struct Dialog {
...
+	} title;
+
+	Color title_color;
+
 	COEVENTS_ARRAY(
 		skip_requested,
 		fadeout_began,
 		fadeout_ended
-	) events;
+	) events, title_events;
```
Akaricchi, line 86: Not sure why `title_events` is needed, doesn't seem like it's used either.

src/dialog.h
```diff
@@ -130,6 +145,9 @@ void dialog_draw(Dialog *d);
 bool dialog_page(Dialog *d)
 	attr_nonnull_all;
 
+void dialog_draw_title(Dialog *d, DialogActor *actor, char *name, char *title);
+	attr_nonnull_all;
```
Akaricchi, line 149:
```
../../../src/dialog.h:149:2: warning: empty declaration
  149 |  attr_nonnull_all;
      |  ^~~~~~~~~~~~~~~~
```

The `attr_nonnull_all` attribute is supposed to be attached to the function declaration; the semicolon on the first line should be removed.

StarWitch:
Changes done!

PR #230, opened 2020-05-22
[spoilers] Small edits to the endings

StarWitch:
Smart quotes were wrong, and I fixed a few small grammar mistakes. No major changes beyond that.

### DIFF
--- a/src/ending.c
+++ b/src/ending.c
@@ -82,7 +82,8 @@ void bad_ending_marisa(Ending *e) {
 	add_ending_entry(e, -1, "It was a ‘Smart Device’ of some kind, like those portable computer-phones from the Outside World.", NULL);
 	add_ending_entry(e, -1, "But this one was far more advanced than anything she’d ever seen.", NULL);
 
-	add_ending_entry(e, -1, "It was almost as if it was made of liquid metal, like mercury, but able to take solid shapes on command. As she pressed its buttons, it cycled through various form factors, such as a thin book, or a handheld phone.", NULL);
+	add_ending_entry(e, -1, "It was almost as if it was made of liquid metal, like mercury, but able to take solid shapes on command.", NULL);
+	add_ending_entry(e, -1, "As she pressed its buttons, it cycled through various form factors, such as a thin book, or a handheld phone.", NULL);
 	add_ending_entry(e, -1, "Indeed, it was more advanced than anything the Outside World was presently capable of, and not in a style even the Lunarians were known for.", NULL);
 
 	add_ending_entry(e, -1, "Once Marisa worked out how to turn it on, the title of a book appeared on its dim display…", NULL);
--- a/src/ending.c
+++ b/src/ending.c
@@ -65,24 +65,25 @@ void bad_ending_marisa(Ending *e) {
 
 	add_ending_entry(e, -1, "Marisa closed the book and let it slam onto the floor.", NULL);
 	add_ending_entry(e, -1, "Patchouli: “Mind your manners! That’s a very old tome!”", NULL);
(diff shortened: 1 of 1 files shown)


src/ending.c
```diff
@@ -65,59 +65,63 @@ void bad_ending_marisa(Ending *e) {
...
-
+	add_ending_entry(e, -1, "Marisa: “It’s so irritatin’! I thought I solved everything but now that tower’s just sittin’ there, ‘n the whole Mountain’s been quarantined…!”", NULL);
 	add_ending_entry(e, -1, "Marisa: “It just ain’t satisfyin’, y’know?”", NULL);
-	add_ending_entry(e, -1, "Flandre: “I could destroy it! Make it go BOOM!”", NULL);
-	add_ending_entry(e, -1, "Marisa: “Gah! Now what did I say about sneakin’ up on people, kiddo?”", NULL);
-	add_ending_entry(e, -1, "Flandre: “Sorry!”", NULL);
+
+	add_ending_entry(e, -1, "Flandre: “I could destroy it, you know.”", NULL);
+	add_ending_entry(e, -1, "Marisa: “Gah! Now what did I say about sneakin’ up on your friends, Flan?”", NULL);
+	add_ending_entry(e, -1, "Flandre: “Sorry, sorry!”", NULL);
```
Akaricchi, line 73: I think I actually prefer the previous version of this; it has more character, I think? The "Make it go BOOM!" line in particular feels very "Flandre" to me, I can easily picture her saying that with exaggerated childish expressiveness. It's endearing.

Do you think there's anything particularly off about this bit that you decided to change it?

src/ending.c
```diff
@@ -65,59 +65,63 @@ void bad_ending_marisa(Ending *e) {
...
 	add_ending_entry(e, -1, "Patchouli: “Thank you for the offer, Flandre, but no.”", NULL);
 	add_ending_entry(e, -1, "Flandre: “Awww…”", NULL);
 
 	add_ending_entry(e, -1, "After finishing tea with the pair of them, Marisa wandered the stacks, looking for inspiration.", NULL);
-	add_ending_entry(e, -1, "Somehow, she tripped on her own feet and landed face-first against a particularly hard and sharp-edged book.", NULL);
-	add_ending_entry(e, -1, "Upon picking it up, she noticed it was actually an ’Intelligent Device’ of some kind, like those portable computers from the Outside World.", NULL);
-	add_ending_entry(e, -1, "This one was far more advanced than anything she’d ever seen.", NULL);
+	add_ending_entry(e, -1, "Somehow, she tripped on her own feet, almost as if something had grabbed her foot, and landed face-first against a particularly hard and sharp-edged book.", NULL);
```
Akaricchi, line 78: Not sure it's a necessary addition, but I'm not really against it either.

I'd suggest maybe
```suggestion
	add_ending_entry(e, -1, "Somehow, she tripped on her own feet, almost as if something had grabbed her, and landed face-first against a particularly hard and sharp-edged book.", NULL);
```
to not repeat feet/foot in the same sentence.

src/ending.c
StarWitch: My main worry was that I'd made her a little TOO childish? It mostly stems from me re-reading her EoSD dialogue and some of the new manga and getting self-conscious of my characterization.

But if you liked it, I can easily put it back. I thought it was cute too.

Akaricchi: Don't be too hard on yourself about characterization. I think it's not really important to keep it super consistent with the canon. I'd say the canon itself isn't always self-consistent, not to mention the countless fanworks.

I see Touhou fanworks as a kind of modern folklore. Being able to reinterpret characters in different ways is a beauty of it. A slightly more childish Flandre hardly compares to the liberties we took with the PC-98 folk, anyway :)

StarWitch: Reverted those changes!

src/ending.c
StarWitch: Revised the entry to be more ~mysterious~.

PR #235, opened 2020-06-08
Various fixes & improvements for concurrent loading

Akaricchi:
* RESF_UNSAFE is removed.
* Resources that don't have to be finalized on the main thread can load completely asynchronously.
* A thread waiting for a concurrent task to complete can start executing that task itself if it hasn't started yet.
* Refactor the resource loading interface, add support for load-time dependencies.
* Main-thread finalization of asynchronously loaded resources is now spread out across multiple frames to mitigate frametime spikes.
* Remove some archaisms from the resource management code.
* Fix potential hashtable synchronization issue.
* Fix some deadlock edge cases.
* Don't spawn more worker threads than there are CPU cores (degrades performance).
* Add `TAISEI_AGGRESSIVE_PRELOAD` env variable to attempt to aggressively discover and preload every possible resource.
* Make `r_texture_fill{,_region}` expect optimal pixmaps, so that it's never forced to convert them on the main thread. The optimal format may be queried with the new `r_texture_optimal_pixmap_format_for_type` API. These functions will also no longer needlessly copy the entire image into a staging buffer - previously they did this even if no conversion was needed.
* Other random changes to facilitate the stuff above.

The overall effect is somewhat faster load times.

Of course it's still all terrible and full of lock contention because I suck at concurrent programming, but it's not worse than it was. Probably.

### DIFF
--- a/src/eventloop/eventloop.c
+++ b/src/eventloop/eventloop.c
@@ -52,6 +52,10 @@ void eventloop_leave(void) {
 	}
 }
 
+FrameTimes eventloop_get_frame_times(void) {
+	return evloop.frame_times;
+}
+
 LogicFrameAction run_logic_frame(LoopFrame *frame) {
 	assert(frame == evloop.stack_ptr);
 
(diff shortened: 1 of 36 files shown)


Akaricchi:
yolo

PR #240, opened 2020-07-24
Texturing overhaul: GPU compression, sRGB sampling, swizzles, etc.

Akaricchi:
~~**Note:** base branch is set to `stage4bg` for a cleaner diff, since this is what I've been testing against. It will be rebased onto master before merging.~~

Support for various GPU compression formats is achieved by integrating the [Basis Universal transcoder](https://github.com/taisei-project/basis_universal). Unfortunately, this is C++ code. I have written a basic C API wrapper for it to avoid introducing C++ into Taisei itself. [The wrapper](https://github.com/taisei-project/basis_universal/blob/taisei/transcoder/basisu_transcoder_c_api.cpp) resides in our basis_universal fork.

Rough TODO:

* [x] (Attempt to) clean up the code, remove debug garbage, etc.
* [x] Split `.basis` loading code into a separate file.
* [ ] Make `.basis` support optional in the engine.
    * [ ] Possibly also transcode `.basis` files into `.png`/`.webp` at build time if support is disabled.
* [x] Utilize basis userdata fields to inform the loader about the nature of the content; needed for some special cases like normalmaps.
* [x] Document some obscure details, like the aforementioned special cases, and the reasoning behind the compressed formats priority list.
* [x] Improve compressed format selection heuristics, in particular for mapping 1- and 2-channel data.
* [x] Improve OpenGL format fuzzy matching heuristics. In particular, teach it to differentiate between classes of formats like unorm, snorm, float, etc.
* [x] Cache transcoded textures on disk to make subsequent loads fast.
* [x] Write some tooling to make creating `.basis` textures more straightforward.
* [x] Test under GLES 3 and GLES 2, ANGLE.
    * [x] GLES 3, Mesa, Linux
    * [x] GLES 3, ANGLE over OpenGL, Linux
    * [x] GLES 3, ANGLE over D3D11, Windows
    * [x] GLES 2, ANGLE over OpenGL, Linux  — works but swizzles are missing
    * [x] ANGLE in WebGL compatibility mode
* [x] Make it compile with Emscripten.
    * [x] And make sure it actually works.
    * **Note:** the relevant texture compression extensions appear to be broken in Chromium 84.0.4147.89 (they fail conformance tests). We could work around this by disabling mipmapping for compressed textures in Chromium browsers, but I'm hoping this will be resolved before release.
* [x] Add "grayscale+alpha" 2-channel mode.
* [x] Account for size limitations for compressed textures.
    * This may differ per implementation; the GLES/WebGL extensions seem more strict.
    * Implemented partially: assumptions about size requirements are hardcoded based on GLES requirements for S3TC and BPTC textures.
* [x] Implement a workaround for missing texture swizzles in GLES2 and WebGL2.
* [x] Hunt down the source files for textures that are currently using lossy webp compression, consider converting them into `.basis` textures.
    * Mostly done; will probably land this separately.
* [ ] Consider a low-quality build profile where all/most textures are compressed (might be useful for WebGL).
* [ ] Write authoring documentation
    * [x] Intro
    * [x] First-time setup
    * [x] TL;DR examples
    * [ ] In-depth encoding nuances
    * [ ] Caveats and limitations
* Probably more stuff that I forgot about.


src/util/miscmath.c
```diff
@@ -495,3 +509,22 @@ uint64_t umuldiv64(uint64_t x, uint64_t multiplier, uint64_t divisor) {
 uint64_t uceildiv64(uint64_t x, uint64_t y) {
 	return x / y + (x % y != 0);
 }
+
+int popcnt32(uint32_t x) {
+	#if TAISEI_BUILDCONF_HAVE_BUILTIN_POPCOUNT
```
jon-turney, line 514: There seems to be an inconsistency here between how this is defined by meson (without a value) in build_config.h, and by src/util/compat.h, which can lead to build failures (e.g. https://travis-ci.com/github/jon-turney/meson-corpus-test/jobs/384217603#L4514)

Akaricchi: Good catch, thanks; should be fixed in 0e25c3a06be326b9b78238b20145003c9ee443da. But I do have to wonder how you managed to trigger this condition in the first place - does the compiler really not have `__builtin_popcount`, or is that a Meson bug?

jon-turney: The compiler I tested with does have  __builtin_popcount.

I think this issue here is that the behaviour of meson when using the `set()` method of a configuration data object with a boolean value is not what you are expecting.

https://mesonbuild.com/Configuration.html#configuring-without-an-input-file

Perhaps you want to use `set10` here instead, depending on what you've decided the meaning of this define is?

https://mesonbuild.com/Reference-manual.html#configuration-data-object

Akaricchi: I'm aware of how the `set` method works, that's what I want. I just accidentally wrote `#if` in place of `#ifdef`.

Disregard my previous comment by the way; I just realized that the error happens when __builtin_popcount *is* found, since the macro just expands into an empty space. It's strange that neither GCC (10.2.1) nor clang (10.0.1) produce this error on my system however, otherwise I definitely should've caught this...

jon-turney: FTR, this is a Ubuntu 20.04 (focal) VM, and the compiler is `gcc (Ubuntu 9.3.0-3ubuntu1) 9.3.0`.

PR #248, opened 2020-08-17
Stage 6 Restructure

StarWitch:
Same as #247. Purely an organizational change, no major restructures, no major variable name changes. Has all of the legacy and jank of the original, but now in bite-sized chunks.


src/stages/stage6/extra.c
```diff
@@ -0,0 +1,41 @@
...
+		*pos += I*VIEWPORT_H;
+		dir += -I;
+	}
+	if(cimag(*pos) > VIEWPORT_H+10) {
+		*pos -= I*VIEWPORT_H;
+		dir += I;
+	}
+
+	return dir;
+}
```
Akaricchi, line 41: This is also only used in Theory of Everything I believe.

src/stages/stage6/extra.h
```diff
@@ -0,0 +1,31 @@
...
+#define SYMMETRYTIME (HIGGSTIME+200)
+#define BREAKTIME (YUKAWATIME+400)
+#define BROGLIE_PERIOD (420 - 30 * global.diff)
+
+#define LASER_EXTENT (4+1.5*global.diff-D_Normal)
+#define LASER_LENGTH 60
+
+#define ELLY_TOE_TARGET_POS (VIEWPORT_W/2+VIEWPORT_H/2*I)
+
+cmplx wrap_around(cmplx*);
```
Akaricchi, line 27: `BROGLIE_PERIOD` is part of de Broglie sign and should go into `spells/broglie.c`. The rest looks like a part of ToE, though some of those macros are used in the ToE spell background function, which is in `elly.c`. I would either move the background implementation into `spells/toe.c` along with all the definitions for now, or move the shared definitions into `elly.h` and the rest into `spells/toe.c`.

src/stages/stage6/extra.h
```diff
@@ -0,0 +1,31 @@
...
+
+int fall_over;
```
Akaricchi, line 29: This one can probably just be folded into `Stage6DrawData` as well. Technically it's part of background animation logic, but as a temporary measure it's fine.

src/stages/stage6/spells/ricci.c
```diff
@@ -0,0 +1,275 @@
...
+#undef SAFE_RADIUS_DELAY
+#undef SAFE_RADIUS_BASE
+#undef SAFE_RADIUS_STRETCH
+#undef SAFE_RADIUS_MIN
+#undef SAFE_RADIUS_MAX
+#undef SAFE_RADIUS_SPEED
+#undef SAFE_RADIUS_PHASE
+#undef SAFE_RADIUS_PHASE_FUNC
+#undef SAFE_RADIUS_PHASE_NORMALIZED
+#undef SAFE_RADIUS_PHASE_NUM
```
Akaricchi, line 272: These can be removed since it's the end of the file anyway.

src/stages/stage6/spells/spells.h
```diff
@@ -0,0 +1,28 @@
...
+ * ---
+ * Copyright (c) 2011-2019, Lukas Weber <laochailan@web.de>.
+ * Copyright (c) 2012-2019, Andrei Alexeyev <akari@taisei-project.org>.
+*/
+
+#ifndef IGUARD_stages_stage6_spells_spells_h
+#define IGUARD_stages_stage6_spells_spells_h
+
+#include "stages/stage6/elly.h"
+#include "taisei.h"
```
Akaricchi, line 13: `taisei.h` must be the first include in all files. Also, don't write/copy-paste the license/include-guard/etc. boilerplate manually. After you add or move/rename a file, simply run `ninja upkeep`. It'll generate them for you.

src/stages/stage6/stage6.c
```diff
@@ -342,8 +147,16 @@ static void stage6_preload(void) {
...
+}
+
 static void stage6_end(void) {
-	stage3d_shutdown(&stage_3d_context);
+	stage6_drawsys_shutdown();
+}
+
+void start_fall_over(void) { //troll
+	fall_over = global.frames;
 }
```
Akaricchi, line 160: This logically belongs in `background_anim.c`.

src/stages/stage6/draw.c
```diff
@@ -0,0 +1,328 @@
...
+}
+
+void stage6_drawsys_init(void) {
+	stage6_draw_data = calloc(1, sizeof(*stage6_draw_data));
+	stage3d_init(&stage_3d_context, 16);
+
+	// TODO: make this background slightly less horribly inefficient
+
+	stage3d_init(&stage_3d_context, 128);
+	stage6_draw_data->fall_over.frames = 0;
```
Akaricchi, line 28: Just tested this. Everything is fine, except there's now a memory leak because `stage3d_init` is called twice here.

Akaricchi (approved):

PR #251, opened 2020-09-01
Switch port fixes

p-sam:
*  Corrected header includes and fs session count to match new libnx (we still don't include the whole <switch.h> to avoid name clashes)
* More env configuration added on debug builds (that helped fix the other issues)
* Game shutdown is now done before __libc_fini_array is called, which avoids
freeing the gl context resources twice (in __libc_fini_array, and in SDL_GL_DeleteContext)
* Linking with libstdc++ is required since the mesa switch portlib was updated

### DIFF
--- a/resources/meson.build
+++ b/resources/meson.build
@@ -112,6 +112,7 @@ if host_machine.system() == 'nx'
                 '--exclude', '**/meson.build',
             ],
             output : shader_pkg_zip,
+            depends : essl_targets,
             depfile : '@0@.d'.format(shader_pkg_zip),
             install : true,
             install_dir : data_path,
--- a/src/arch_switch.c
+++ b/src/arch_switch.c
@@ -78,11 +78,6 @@ void userAppInit(void) {
 
 attr_used
 void userAppExit(void) {
-	if(g_nxAtExitFn != NULL) {
-		NX_LOG("calling exit callback");
-		g_nxAtExitFn();
-		g_nxAtExitFn = NULL;
-	}
 	socketExit();
 	appletUnlockExit();
 }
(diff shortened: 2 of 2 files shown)


resources/meson.build
```diff
@@ -111,6 +111,7 @@ if host_machine.system() == 'nx'
                 '--exclude', '**/*.spv',
                 '--exclude', '**/meson.build',
             ],
+            input : essl_targets,
```
Akaricchi, line 114: I don't think this is the right way to do this; `input:` is for inputs used on the commandline via the `@INPUT@` placeholder(s). You want `depends:` instead.

switch/crossfile.sh
```diff
@@ -18,7 +18,8 @@ function bin_path() {
 	which "${TOOL_PREFIX}$1"
 }
 
-ADDITIONAL_LINK_FLAGS="-specs=$DEVKITPRO/libnx/switch.specs"
+# Link with libstdc++ since libEGL requires it
+ADDITIONAL_LINK_FLAGS="-specs=$DEVKITPRO/libnx/switch.specs -lstdc++"
```
Akaricchi, line 22: Is this really necessary? I'm pretty sure that since #240 Taisei already depends on a C++ standard library, because the basis transcoder is C++ and uses some STL stuff. Meson should automatically use a C++ compiler (e.g. `g++`) at the linking stage because of this, which I think already should pull in the C++ stdlib. Does that not work as is?

resources/meson.build
p-sam: Replaced `input `with `depends`, the archive built is not empty as expected (even on a clean build)

switch/crossfile.sh
p-sam: You're right it's not needed since the texture PR, this was done prior, and I removed it

Akaricchi (approved):

PR #259, opened 2020-11-08
Stop vendoring cglm

Akaricchi:
* Remove cglm from submodules
* Add cglm to list of dependencies
* Prefer system version of cglm, with a wrap fallback (as with other dependencies)


Akaricchi:
Waiting at least until cglm 0.8.0 before landing this.

lanodan:
cglm 0.8.0 got released few days ago btw

PR #263, opened 2020-12-06
Title Cards

StarWitch:
Moving over from #220.

### DIFF
--- a/src/dialog.c
+++ b/src/dialog.c
@@ -104,7 +104,7 @@ void dialog_update(Dialog *d) {
 		fapproach_asymptotic_p(&a->focus, a->target_focus, 0.12, 1e-3);
 	}
 
-	if (d->title.active) {
+	if(d->title.active) {
 		if(d->title.timeout > 0) {
 			fapproach_asymptotic_p(&d->title.box.opacity, 1, 0.05, 1e-3);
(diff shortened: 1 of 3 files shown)


src/dialog.c
```diff
@@ -103,6 +103,24 @@ void dialog_update(Dialog *d) {
 
 		fapproach_asymptotic_p(&a->focus, a->target_focus, 0.12, 1e-3);
 	}
+
+	if (d->title.active) {
```
laochailan, line 107: should this be done with tasks? @Akaricchi 
like common_easing_animate!

Akaricchi: It can be, but I don't really know if it should be. The logic here isn't particularly convoluted without tasks, so I don't have a strong opinion here. Replacing the entirety of `dialog_update` with a task might be good I suppose, so that stage.c wouldn't have to call it explicitly.

Akaricchi: By the way, extraneous space after `if` on this line.

laochailan (commented):
looks good to me!

Akaricchi:
Merged an edited version of this in 5a0a236ff1cb5140513d1a0c2ce6e5747fb5e28e
The code is based on yours, so I've tagged you as a co-author.

PR #266, opened 2020-12-16
Build testing with Actions

StarWitch:
Basic build testing. Not much to it.

1. Setups up a container (Ubuntu and macOS)
2. Installs all the tools it needs (with `apt`, `pip`, and/or `brew`)
3. Compiles Taisei with (with code-health options like `-Dwerror`)
4. Runs a 1-second ReimuA `.tsr` replay file to ensure the game boots at all
5. Outputs the results for download on GitHub as an "artifact" if there were any errors

No Windows build environment since we cross-compile on Linux anyways (and compiling on Windows is a Challenge).


.github/workflows/main.yml
```diff
@@ -0,0 +1,82 @@
...
+      run: git submodule update --init --recursive
+
+    - name: Install Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+
+    # needed because apt version is too old
+    - name: Install Build Tools (pip)
+      run: pip install meson ninja
```
Akaricchi, line 36: Ubuntu 20.04 should have meson 0.53.2, which should be new enough... if that doesn't work, we need to fix our build system.

.github/workflows/main.yml
```diff
@@ -0,0 +1,82 @@
...
+      - "**.c"
+      - "**.h"
+      - "**.yml"
+      - "**.yaml"
+  pull_request:
+    paths:
+      - "**.c"
+      - "**.h"
+      - "**.yml"
+      - "**.yaml"
```
Akaricchi, line 17: Might want to add `*.build` as well.

.github/workflows/main.yml
```diff
@@ -0,0 +1,82 @@
...
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+
+    # needed because apt version is too old
+    - name: Install Build Tools (pip)
+      run: pip install meson ninja
+
+    - name: Install Build Tools (apt)
+      run: sudo apt install cmake build-essential libsdl2-dev libsdl2-mixer-dev libogg-dev libopusfile-dev libpng-dev libzip-dev libx11-dev libwayland-dev imagemagick python3-docutils libwebp-dev
```
Akaricchi, line 39: `libsdl2-mixer-dev` shouldn't be needed.

.github/workflows/main.yml
```diff
@@ -0,0 +1,82 @@
...
+
+    - name: Configure (Meson)
+      run: meson setup build/ -Dinstall_relative=true -Dbuildtype=debug -Db_ndebug=false -Db_lto=false -Dstrip=false -Ddeveloper=true
```
Akaricchi, line 42: Maybe it makes more sense to do a release build here, since that is the least tested configuration during development. We also disable precompiled headers on our OBS builds for that reason, to make sure nothing breaks without them. Not sure why the `install_relative`, but it probably doesn't matter much if we aren't publishing the build artifacts.

Akaricchi: Maybe also add `--wrap-mode=nofallback` to ensure system dependencies are used.

Akaricchi: I'd also do `-Dwerror=true -Ddeprecation_warnings=ignore`

StarWitch: `-Dwerror=true` will make many, many parts of the code fail due to warnings, including submodules like `basisu`. Is that something we really want for a CI pipeline?

Ex:

```
FAILED: src/25a6634@@taisei@exe/renderer_glcommon_opengl.c.o 
cc -Isrc/25a6634@@taisei@exe -Isrc -I../src -I../subprojects/basis_universal/transcoder -Isubprojects/cglm/include -I../subprojects/cglm/include -I../subprojects/koishi/include -I../subprojects/glad/include -I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/SDL2 -I/usr/include/opus -fdiagnostics-color=always -DNDEBUG -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Werror -std=gnu11 -O3 -D_REENTRANT -DBASISD_IS_BIG_ENDIAN=0 -DBASISD_USE_UNALIGNED_WORD_READS=0 -DBASISD_SUPPORT_ASTC=1 -DBASISD_SUPPORT_ATC=1 -DBASISD_SUPPORT_BC7=1 -DBASISD_SUPPORT_BC7_MODE5=1 -DBASISD_SUPPORT_DXT1=1 -DBASISD_SUPPORT_DXT5A=1 -DBASISD_SUPPORT_ETC2_EAC_A8=1 -DBASISD_SUPPORT_ETC2_EAC_RG11=1 -DBASISD_SUPPORT_FXT1=0 -DBASISD_SUPPORT_PVRTC1=0 -DBASISD_SUPPORT_PVRTC2=1 -DBASISD_SUPPORT_UASTC=1 -Wall -Wpedantic -Werror=implicit-function-declaration -Werror=incompatible-pointer-types -Wabsolute-value -Wcast-align -Wcast-align=strict -Wcast-function-type -Wclobbered -Wduplicated-branches -Wduplicated-cond -Wformat-security -Wignored-qualifiers -Winit-self -Wlogical-op -Wmissing-prototypes -Wno-implicit-fallthrough -Wno-long-long -Wno-missing-braces -Wnull-dereference -Wparentheses -Wshadow=compatible-local -Wstrict-prototypes -Wtype-limits -Wunreachable-code -fexcess-precision=standard -fmerge-all-constants -fno-math-errno -fno-signaling-nans -fno-trapping-math -Wno-deprecated-declarations -msse -mfpmath=sse -fpch-preprocess -include taisei_pch.h -MD -MQ 'src/25a6634@@taisei@exe/renderer_glcommon_opengl.c.o' -MF 'src/25a6634@@taisei@exe/renderer_glcommon_opengl.c.o.d' -o 'src/25a6634@@taisei@exe/renderer_glcommon_opengl.c.o' -c ../src/renderer/glcommon/opengl.c
In file included from ../src/hashtable.inc.h:8,
                 from ../src/hashtable_incproxy.inc.h:10,
                 from ../src/hashtable_predefs.inc.h:31,
                 from ../src/hashtable.h:77,
                 from ../src/util/kvparser.h:16,
                 from ../src/util.h:22,
                 from ../src/global.h:17,
                 from ../src/pch/taisei_pch.h:14:
In function ‘glcommon_ext_flag’,
    inlined from ‘glcommon_check_extension’ at ../src/renderer/glcommon/opengl.c:81:20:
../src/log.h:132:34: error: ‘%s’ directive argument is null [-Werror=format-overflow=]
  132 |  #define _do_log(func, lvl, ...) (func)(lvl, __func__, "<unknown>", 0, __VA_ARGS__)
      |                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/log.h:142:24: note: in expansion of macro ‘_do_log’
  142 | #define log_fatal(...) _do_log(_taisei_log_fatal, LOG_FATAL, __VA_ARGS__)
      |                        ^~~~~~~
../src/renderer/glcommon/opengl.c:55:3: note: in expansion of macro ‘log_fatal’
   55 |   log_fatal("Bad extension string: %s", ext);
      |   ^~~~~~~~~
../src/renderer/glcommon/opengl.c: In function ‘glcommon_check_extension’:
../src/renderer/glcommon/opengl.c:55:36: note: format string is defined here
   55 |   log_fatal("Bad extension string: %s", ext);
      |                                    ^~
cc1: all warnings being treated as errors
[39/284] Compiling C object 'src/25a6634@@taisei@exe/boss.c.o'.
[40/284] Compiling C++ object 'subprojects/basis_universal/1d967b7@@basisu_transcoder@sta/transcoder_basisu_transcoder.cpp.o'.
FAILED: subprojects/basis_universal/1d967b7@@basisu_transcoder@sta/transcoder_basisu_transcoder.cpp.o 
c++ -Isubprojects/basis_universal/1d967b7@@basisu_transcoder@sta -Isubprojects/basis_universal -I../subprojects/basis_universal -fvisibility=hidden -fdiagnostics-color=always -DNDEBUG -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wnon-virtual-dtor -Werror -std=c++14 -fno-exceptions -fno-rtti -O3 -fPIC -Wall -Wno-class-memaccess -Wno-misleading-indentation -Wno-sign-compare -Wno-unused-but-set-variable -Wno-unused-variable -fno-exceptions -fno-rtti -fno-strict-aliasing -DBASISD_IS_BIG_ENDIAN=0 -DBASISD_USE_UNALIGNED_WORD_READS=0 -DBASISD_SUPPORT_ASTC=1 -DBASISD_SUPPORT_ATC=1 -DBASISD_SUPPORT_BC7=1 -DBASISD_SUPPORT_BC7_MODE5=1 -DBASISD_SUPPORT_DXT1=1 -DBASISD_SUPPORT_DXT5A=1 -DBASISD_SUPPORT_ETC2_EAC_A8=1 -DBASISD_SUPPORT_ETC2_EAC_RG11=1 -DBASISD_SUPPORT_FXT1=0 -DBASISD_SUPPORT_PVRTC1=0 -DBASISD_SUPPORT_PVRTC2=1 -DBASISD_SUPPORT_UASTC=1 -MD -MQ 'subprojects/basis_universal/1d967b7@@basisu_transcoder@sta/transcoder_basisu_transcoder.cpp.o' -MF 'subprojects/basis_universal/1d967b7@@basisu_transcoder@sta/transcoder_basisu_transcoder.cpp.o.d' -o 'subprojects/basis_universal/1d967b7@@basisu_transcoder@sta/transcoder_basisu_transcoder.cpp.o' -c ../subprojects/basis_universal/transcoder/basisu_transcoder.cpp
../subprojects/basis_universal/transcoder/basisu_transcoder.cpp: In function ‘bool basist::transcode_uastc_to_bc7(const basist::unpacked_uastc_block&, basist::bc7_optimization_results&)’:
../subprojects/basis_universal/transcoder/basisu_transcoder.cpp:12710:26: error: ‘*((void*)& best_pbits +4)’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
12710 |    dst_blk.m_pbits[0][1] = best_pbits[1];
      |    ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~
../subprojects/basis_universal/transcoder/basisu_transcoder.cpp:12709:26: error: ‘best_pbits’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
12709 |    dst_blk.m_pbits[0][0] = best_pbits[0];
      |    ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
ninja: build stopped: subcommand failed.
```

.github/workflows/main.yml
StarWitch: Yep, missed that. It was from before when I was testing with Ubuntu 18.04, which had an old `meson`. I bumped it up to 20.04 since that's going to become the GitHub standard in a month anyways.

.github/workflows/main.yml
Akaricchi: Meson has gained an ability to set `werror` per-subproject, though I'm not sure since when. If configuring initially with `werror=true` defaults it to true for subprojects as well, then we should be able to set it just for the main project as a separate step, e.g.
```
$ meson setup -Dfoo=bar build/
$ meson configure build/ -Dwerror=true
$ meson compile -C build/
```

Alternatively, you should be able to disable werror for specific subprojects manually:
```
$ meson setup -Dfoo=bar -Dwerror=true -Dbasis_universal:werror=false build/
$ meson compile -C build/
```

If neither of this works, then meson is too old. Ideally we should be testing against our minimum supported version, but I'd like to bump that requirement to something more reasonable anyway.

As for warnings that occur in the main project, they must be just fixed (or silenced, in cases where the compiler is being dumb).

.github/workflows/main.yml
```diff
@@ -31,15 +33,11 @@ jobs:
       with:
         python-version: '3.x'
 
-    # needed because apt version is too old
-    - name: Install Build Tools (pip)
-      run: pip install meson ninja
-
     - name: Install Build Tools (apt)
-      run: sudo apt install cmake build-essential libsdl2-dev libsdl2-mixer-dev libogg-dev libopusfile-dev libpng-dev libzip-dev libx11-dev libwayland-dev imagemagick python3-docutils libwebp-dev
+      run: sudo apt install meson cmake build-essential libsdl2-dev libogg-dev libopusfile-dev libpng-dev libzip-dev libx11-dev libwayland-dev imagemagick python3-docutils libwebp-dev
```
Akaricchi, line 37: cmake shouldn't be needed either, missed it last time.

Akaricchi: Or imagemagick. It's only used for development-specific tasks like atlas generation, but not during build.

.github/workflows/main.yml
```diff
@@ -31,15 +33,11 @@ jobs:
...
 
     - name: Configure (Meson)
-      run: meson setup build/ -Dinstall_relative=true -Dbuildtype=debug -Db_ndebug=false -Db_lto=false -Dstrip=false -Ddeveloper=true
+      run: meson setup --wrap-mode=nofallback build/ -Db_lto=false -Dstrip=false -Ddeprecation_warnings=ignore
```
Akaricchi, line 40: Sorry, I actually meant to say `-Ddeprecation_warnings=no-error`, not `ignore`.

.github/workflows/main.yml
StarWitch: Got it. Running it on my own private fork, it seems that `src/coroutines.c` has some hard warnings in it:

```
../src/coroutine.c:150:35: error: statement with no effect [-Werror=unused-value]
 #define STAT_VAL_SET(name, value) (0)
```

macOS also fails with a different warning.

You can see here: https://github.com/StarWitch/taisei/pull/1/checks?check_run_id=1574067166

If we want, we could have two runners: one that's more permissive for most things ("this actually compiles"), and one that hard-fails on these sorts of warning-errors ("code health"). The runners don't take too long so I don't think we're at a risk of running out of free CI minutes.

StarWitch: All fixed in #273.

.github/workflows/main.yml
```diff
@@ -0,0 +1,104 @@
+name: taisei-build-tests
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - v1.*
```
Akaricchi, line 7: We also have `v1.*.x` branches (not tags) for staging stable releases before they are tagged.

.github/workflows/main.yml
```diff
@@ -0,0 +1,104 @@
...
+    paths:
+      - "**.c"
+      - "**.h"
+      - "**.yml"
+      - "**.yaml"
+      - "**.build"
+
+jobs:
+  linux-test-build:
+    if: "!contains(github.event.head_commit.message, '[skip ci]')"
```
Akaricchi, line 24: What happens if a push contains several commits, and only one of them has `[skip ci]`? Ideally we want to run CI unless *all* of the pushed commits have the skip tag.

StarWitch: `*.head_commit.*` only looks at the top-most `HEAD` commit, rather than the entire stack of commits. So any commit after a commit with `[skip ci]` will still run so long as it doesn't contain that string.

.github/workflows/main.yml
StarWitch: Added!

.github/workflows/main.yml
StarWitch: One thing to note, actually, is that `github.events.*` only apply to non-PR actions, so commits with `[skip ci]` will still run in PRs. A bit annoying but if you're putting up a PR it should probably run every time.

Akaricchi (approved):
I guess we can finally merge it and see how well it does. Ultimately there are some improvements to be made (such as using meson's test suite feature instead of hardcoding the replay command here), but we can iron those out later.

PR #273, opened 2020-12-24
More compiler warning fixes

StarWitch:
Not much to it. Played whack-a-mole until the pipeline on my fork passed with `-Dwerrors=true`. See: https://github.com/StarWitch/taisei/pull/1/checks?check_run_id=1603187734

Potentially some bugs squashed in the process as well.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -29,7 +29,7 @@ project('taisei', 'c',
     ]
 )
 
-minimum_recommended_meson_version = '0.49.0'
+minimum_recommended_meson_version = '0.53.0'
 
 if meson.version().version_compare('<@0@'.format(minimum_recommended_meson_version))
     warning('Old Meson version detected. Try upgrading to at least @0@ if the build fails.'.format(minimum_recommended_meson_version))
--- a/meson.build
+++ b/meson.build
@@ -29,12 +29,6 @@ project('taisei', 'c',
     ]
 )
 
-minimum_recommended_meson_version = '0.53.0'
-
-if meson.version().version_compare('<@0@'.format(minimum_recommended_meson_version))
-    warning('Old Meson version detected. Try upgrading to at least @0@ if the build fails.'.format(minimum_recommended_meson_version))
-endif
-
 is_debug_build = get_option('debug')
 is_developer_build = (get_option('developer') == 'auto' ? is_debug_build : get_option('developer') == 'true')
 
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 project('taisei', 'c',
     license : 'MIT',
     version : 'v1.4-dev',
-    meson_version : '>=0.48.0',
+    meson_version : '>=0.53.0',
     default_options : [
         # Should really be c11, but gnu11 is a safer default because everything is terrible.
         'c_std=gnu11',
@@ -39,7 +39,7 @@ is_debug_build = get_option('debug')
 is_developer_build = (get_option('developer') == 'auto' ? is_debug_build : get_option('developer') == 'true')
 
 cc = meson.get_compiler('c')
-python3 = import('python3').find_python()
+python = import('python').find_installation()
 macos_app_bundle = get_option('macos_bundle') and host_machine.system() == 'darwin'
 
 subdir('scripts')
@@ -85,6 +85,7 @@ taisei_c_args = [
     '-Wparentheses',
     '-Wshadow=compatible-local',
     '-Wsometimes-uninitialized',
+    '-Wstrict-overflow=0',
     '-Wstrict-prototypes',
     '-Wtype-limits',
     '-Wunneeded-internal-declaration',
@@ -283,7 +284,7 @@ if macos_app_bundle
 
     # arguments must be strings...
     meson.add_install_script(
-        python3.path(),
+        python.path(),
         join_paths(meson.source_root(), 'scripts', 'macos-install-dylibs.py'),
         ':'.join(meson.get_cross_property('macos_lib_path', [])),
         ':'.join(meson.get_cross_property('macos_tool_path', [])),
--- a/src/plrmodes/marisa_b.c
+++ b/src/plrmodes/marisa_b.c
@@ -104,12 +104,10 @@ TASK(marisa_star_slave_projectile, {
 }) {
 	MarisaBController *ctrl = ARGS.ctrl;
 	Player *plr = ctrl->plr;
-	MarisaBSlave *slave = ENT_UNBOX(ARGS.slave);
+	MarisaBSlave *slave = NOT_NULL(ENT_UNBOX(ARGS.slave));
 
-	if(slave) {
-		if(!slave->alive || !player_should_shoot(plr)) {
-			return;
-		}
+	if(!slave->alive || !player_should_shoot(plr)) {
+		return;
 	}
 
 	Projectile *p = TASK_BIND(PROJECTILE(
--- a/src/plrmodes/marisa_b.c
+++ b/src/plrmodes/marisa_b.c
@@ -104,9 +104,12 @@ TASK(marisa_star_slave_projectile, {
 }) {
 	MarisaBController *ctrl = ARGS.ctrl;
 	Player *plr = ctrl->plr;
+	MarisaBSlave *slave = ENT_UNBOX(ARGS.slave);
 
-	if(!ENT_UNBOX(ARGS.slave)->alive || !player_should_shoot(plr)) {
-		return;
+	if(slave) {
+		if(!slave->alive || !player_should_shoot(plr)) {
+			return;
+		}
 	}
 
 	Projectile *p = TASK_BIND(PROJECTILE(
@@ -126,7 +129,7 @@ TASK(marisa_star_slave_projectile, {
 	cmplx next_pos = p->pos;
 
 	for(int t = 0;; ++t) {
-		MarisaBSlave *slave = ENT_UNBOX(ARGS.slave);
+		slave = ENT_UNBOX(ARGS.slave);
 
 		if(slave == NULL || !slave->alive || !player_should_shoot(plr)) {
 			break;
--- a/src/plrmodes/youmu_b.c
+++ b/src/plrmodes/youmu_b.c
@@ -229,7 +229,7 @@ TASK(youmu_haunting_shot_homing, { YoumuBController *ctrl; }) {
 }
 
 TASK(youmu_orb_homing_spirit_expire, { BoxedProjectile p; }) {
-	Projectile *p = ENT_UNBOX(ARGS.p);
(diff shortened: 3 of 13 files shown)


src/stages/stage2/spells/amulet_of_harm.c
```diff
@@ -50,8 +50,8 @@ TASK(amulet_fire_spinners, { BoxedEnemy core; BoxedProjectileArray *spinners; })
 
 		ENT_ARRAY_FOREACH(ARGS.spinners, Projectile *p, {
 			int cnt = difficulty_value(12, 16, 22, 24);
-			for(int i = 0; i < cnt; ++i) {
-				cmplx ca = circle_dir(i, cnt);
+			for(int x = 0; x < cnt; ++x) {
+				cmplx ca = circle_dir(x, cnt);
```
StarWitch, line 54: For reference, `gcc` complained that `int i = 0` (etc) was already used above, and it wanted a different var name.

src/stages/stage2/timeline.c
```diff
@@ -80,19 +80,14 @@ TASK(spinshot_fairy_attack, {
...
-		e = ENT_UNBOX(ARGS.e);
-
-		if(e) {
-			ref_pos = e->pos;
-		}
+		e = NOT_NULL(ENT_UNBOX(ARGS.e));
 
 		ENT_ARRAY_FOREACH_COUNTER(&projs, int i, Projectile *p, {
-			cmplx ofs = ref_pos - proj_origins[i];
+			cmplx ofs = e->pos - proj_origins[i];
```
StarWitch, line 90: Seemed redundant to have a separate var, but if there was a specific reason for it, it can get put back.

src/stages/stage3_events.c
```diff
@@ -250,7 +250,7 @@ static int stage3_burstfairy(Enemy *e, int t) {
 
 	charge_effect(e, t, bursttime);
 
-	FROM_TO_SND("shot1_loop", bursttime, bursttime + burstspan, step) {
+	FROM_TO_SND("shot1_loop", bursttime, bursttime + burstspan * step - 1, step) {
```
StarWitch, line 253: Not sure about this one. Seems fine during testing???

src/plrmodes/marisa_b.c
```diff
@@ -104,9 +104,12 @@ TASK(marisa_star_slave_projectile, {
...
 	MarisaBController *ctrl = ARGS.ctrl;
 	Player *plr = ctrl->plr;
+	MarisaBSlave *slave = ENT_UNBOX(ARGS.slave);
 
-	if(!ENT_UNBOX(ARGS.slave)->alive || !player_should_shoot(plr)) {
-		return;
+	if(slave) {
+		if(!slave->alive || !player_should_shoot(plr)) {
+			return;
+		}
```
Akaricchi, line 112: This one can/should just be `NOT_NULL(ENT_UNBOX(ARGS.slave))`. That'll throw an assertion error if the task happens to be called without a slave reference.

Note that this doesn't apply to the `ENT_UNBOX` inside the for loop later in this function.

src/plrmodes/youmu_b.c
```diff
@@ -246,7 +246,10 @@ TASK(youmu_orb_homing_spirit_expire, { BoxedProjectile p; }) {
 }
 
 static int youmu_orb_homing_spirit_timeout(Projectile *orb) {
-	return orb->timeout - projectile_time(orb);
+	if(orb) {
+		return orb->timeout - projectile_time(orb);
+	}
+	return 0;
 }
```
Akaricchi, line 253: Adding `attr_nonnull(1)` before the function declaration would be a better idea - this should never be called with a NULL argument.

src/stages/stage2/timeline.c
Akaricchi: This is incorrect. `ENT_UNBOX(ARGS.e)` *may* be `NULL` here, hence the check in the original code. This happens if the fairy dies before finishing the attack. The spawner task dies with the fairy, but the attack logic needs to keep running here to process the already spawned bullets. `ref_pos` is used to remember the last position of the fairy before it died.

There seems to be nothing wrong with the original code - does it actually trip any warnings?

src/stages/stage3_events.c
Akaricchi: What is this change trying to solve?

Akaricchi (dismissed):

src/stages/stage3_events.c
StarWitch:
```
../src/stages/stage3_events.c: In function ‘stage3_burstfairy’:
../src/stage.h:35:101: error: assuming signed overflow does not occur when assuming that (X + c) >= X is always true [-Werror=strict-overflow]
 #define FROM_TO(start,end,step) _ni = _ni; _i = (*__timep - (start))/(step); if(*__timep >= (start) && *__timep <= (end) && !((*__timep - (start)) % (step)))
                                                                                 ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~
../src/stage.h:58:35: note: in expansion of macro ‘FROM_TO’
 #define PLAY_FOR(name,start, end) FROM_TO(start,end,2) { play_sfx_loop(name); }
                                   ^~~~~~~
../src/stage.h:61:41: note: in expansion of macro ‘PLAY_FOR’
 #define FROM_TO_SND(snd,start,end,step) PLAY_FOR(snd,start,end); FROM_TO(start,end,step)
                                         ^~~~~~~~
../src/stages/stage3_events.c:253:2: note: in expansion of macro ‘FROM_TO_SND’
  FROM_TO_SND("shot1_loop", bursttime, bursttime + burstspan, step) {
  ^~~~~~~~~~~
```

src/stages/stage2/timeline.c
StarWitch: All of the fixes in the branch were directly to fix compiler errors.

Here's the log:

```
 ../src/stages/stage2/timeline.c: In function ‘COTASKTHUNK_spinshot_fairy_attack’:
../src/stages/stage2/timeline.c:95:10: error: ‘__imag__ ref_pos’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
    cmplx ofs = ref_pos - proj_origins[i];
          ^~~
../src/stages/stage2/timeline.c:83:8: note: ‘__imag__ ref_pos’ was declared here
  cmplx ref_pos;
        ^~~~~~~
../src/stages/stage2/timeline.c:95:10: error: ‘__real__ ref_pos’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
    cmplx ofs = ref_pos - proj_origins[i];
          ^~~
```

src/stages/stage3_events.c
Akaricchi: I think we should just disable this warning globally. It's mostly useless and has lots of false positives. 

The changed line needs to be reverted, since it alters behavior.

src/stages/stage2/timeline.c
Akaricchi: This should be fixed by initializing `ref_pos`

```c
cmplx ref_pos = NOT_NULL(ENT_UNBOX(ARGS.e))->pos;
```

`ARGS.e` is expected to be a valid reference when the task starts, so this is fine.

src/plrmodes/youmu_b.c
StarWitch: I added `attr_nonnull(1)` and the compiler still spits out the same warning.

https://github.com/StarWitch/taisei/pull/1/checks?check_run_id=1606225668

https://github.com/StarWitch/taisei/pull/1/commits/5ffa51b633bf3dc94667008dc9798c314c0e0a61#diff-505af192e6781de8f60b076c8d9095e18d58d47b771836184907aa22f7efefc2R248

src/stages/stage3_events.c
StarWitch: Disabled it globally with `-Wstrict-overflow=0`.

src/plrmodes/youmu_b.c
Akaricchi: The compiler sucks, then. `NOT_NULL(orb)` should shut it up, or `assume(orb != NULL);` before the return statement.

meson.build
```diff
@@ -1,7 +1,7 @@
 project('taisei', 'c',
     license : 'MIT',
     version : 'v1.4-dev',
-    meson_version : '>=0.48.0',
+    meson_version : '>=0.53.0',
```
Akaricchi, line 4: Since this has been bumped, we should also remove the `minimum_recommended_meson_version` code just below the `project` statement, and update the README.

src/stages/stage2/timeline.c
```diff
@@ -80,16 +80,9 @@ TASK(spinshot_fairy_attack, {
...
-	Enemy *e;
-
 	for(int t = 0; t < ARGS.activate_time; ++t, YIELD) {
 		int live_count = 0;
-		e = ENT_UNBOX(ARGS.e);
-
-		if(e) {
-			ref_pos = e->pos;
-		}
+		cmplx ref_pos = NOT_NULL(ENT_UNBOX(ARGS.e))->pos;
```
Akaricchi, line 85: Not like that. The code inside the `for` loop must not assume that `ENT_UNBOX(ARGS.e)` isn't NULL, it has to check. It is guaranteed to not be null only on the very first iteration (after that we `YIELD`, and the entity may die before the task is re-entered; this is a valid state). GCC doesn't know that the first `ENT_UNBOX` is a guaranteed non-null, hence the maybe-uninitialized warning. 

Here's what needs to be done:

* Code inside the loop has to be reverted to its pre-PR state; it is correct.
* `cmplx ref_pos = NOT_NULL(ENT_UNBOX(ARGS.e))->pos` needs to go before the loop.
    * Technically `ref_pos` could be initialized to anything, e.g. 0. `NOT_NULL(ENT_UNBOX(ARGS.e))->pos` just adds an extra sanity check (to ensure that we didn't call the task with an invalid `ARGS.e` in the first place), and establishes that this variable is meant to track position of `ARGS.e`.

Akaricchi (changes requested):

meson.build
```diff
@@ -29,7 +29,7 @@ project('taisei', 'c',
     ]
 )
 
-minimum_recommended_meson_version = '0.49.0'
+minimum_recommended_meson_version = '0.53.0'
 
 if meson.version().version_compare('<@0@'.format(minimum_recommended_meson_version))
     warning('Old Meson version detected. Try upgrading to at least @0@ if the build fails.'.format(minimum_recommended_meson_version))
```
Akaricchi, line 35: All this stuff can just be removed. Meson will abort if it's older than `0.53.0` already.

Akaricchi (approved):

PR #278, opened 2021-02-20
Dockerfiles

StarWitch:
Some Dockerfiles to help with build images. One is only currently in use: `switch-toolkit`. The others could potentially get used if we decide the CI speedup is worth it.


StarWitch:
Got rid of the abomination that was the ANGLE Docker build. It's not needed anymore since we have GHA working nicely. Should make this PR much easier to review.

StarWitch:
Merging this as it's low-impact. We can fix any remaining issues later.

PR #280, opened 2021-03-01
Stage 5 Remaster

StarWitch:
Modernizing Stage 5. WIP so commits can be commented on.

### DIFF
--- a/src/stages/stage5/iku.c
+++ b/src/stages/stage5/iku.c
@@ -14,8 +14,6 @@
 
 #include "common_tasks.h"
 
-MODERNIZE_THIS_FILE_AND_REMOVE_ME
-
 static void iku_slave_draw(EntityInterface *e) {
 	IkuSlave *slave = ENT_CAST(e, IkuSlave);
 	r_draw_sprite(&(SpriteParams) {
@@ -26,19 +24,16 @@ static void iku_slave_draw(EntityInterface *e) {
 	});
 }
 
-void iku_lightning_particle(cmplx pos, int t) {
-	if(!(t % 5)) {
-		char *part = frand() > 0.5 ? "lightning0" : "lightning1";
-		PARTICLE(
-			.sprite = part,
-			.pos = pos,
-			.color = RGBA(1.0, 1.0, 1.0, 0.0),
-			.timeout = 20,
-			.draw_rule = Fade,
-			.flags = PFLAG_REQUIREDPARTICLE,
-			.angle = frand()*2*M_PI,
-		);
-	}
+void iku_lightning_particle(cmplx pos) {
+	char *part = rng_chance(0.5) ? "lightning0" : "lightning1";
+	PARTICLE(
+		.sprite = part,
+		.pos = pos,
+		.color = RGBA(1.0, 1.0, 1.0, 0.0),
+		.timeout = 20,
+		.draw_rule = pdraw_timeout_fade(0.8, 0.0),
+		.angle = rng_real() * M_TAU,
+	);
 }
 
 TASK(iku_slave_visual, { BoxedIkuSlave slave; }) {
@@ -50,6 +45,7 @@ TASK(iku_slave_visual, { BoxedIkuSlave slave; }) {
 	for(;;WAIT(period)) {
 		RNG_ARRAY(rand, 2);
 		cmplx offset = vrng_sreal(rand[0]) * 5 + vrng_sreal(rand[1]) * 5 * I;
+		iku_lightning_particle(slave->pos);
 		PARTICLE(
 			.sprite_ptr = slave->sprites.cloud,
 			.pos = slave->pos,
@@ -92,56 +88,21 @@ DEFINE_EXTERN_TASK(iku_slave_move) {
 	}
 }
 
-
 void iku_nonspell_spawn_cloud(void) {
-	tsrand_fill(4);
-	float v = (afrand(2)+afrand(3))*0.5+1.0;
+	RNG_ARRAY(rand, 4);
+	float v = (vrng_sreal(rand[2]) + vrng_sreal(rand[3])) * 0.5 + 1.0;
 
 	PROJECTILE(
 		// FIXME: add prototype, or shove it into the basic ones somehow,
 		// or just replace this with some thing else
-		.sprite_ptr = get_sprite("part/lightningball"),
-		.size = 48 * (1+I),
-		.collision_size = 21.6 * (1+I),
-
-		.pos = VIEWPORT_W*afrand(0)-15.0*I,
+		.sprite_ptr = res_sprite("part/lightningball"),
+		.size = 48 * (1 + I),
+		.collision_size = 21.6 * (1 + I),
+		.pos = VIEWPORT_W * vrng_sreal(rand[0]) - 15.0 * I,
 		.color = RGBA_MUL_ALPHA(0.2, 0.0, 0.4, 0.6),
-		.rule = accelerated,
-		.args = {
-			1-2*afrand(1)+v*I,
-			-0.01*I
-		},
+		.move = move_accelerated(1 - 2 * vrng_sreal(rand[1]) + v * I, -0.01 * I),
 		.shader = "sprite_default",
 	);
-}
-
-static cmplx induction_bullet_traj(Projectile *p, float t) {
-	return p->pos0 + p->args[0]*t*cexp(p->args[1]*t);
-}
-
-int iku_induction_bullet(Projectile *p, int time) {
-	if(time < 0) {
-		return ACTION_ACK;
-	}
-
-	float t = time*sqrt(global.diff);
-
-	if(global.diff > D_Normal && !p->args[2]) {
-		t = time*0.6;
-		t = 230-t;
-		if(t < 0)
-			return ACTION_DESTROY;
-	}
-
-	p->pos = induction_bullet_traj(p,t);
-
-	if(time == 0) {
-		// don't lerp; the spawn position is very different on hard/lunatic and would cause false hits
-		p->prevpos = p->pos;
-	}
-
-	p->angle = carg(p->args[0]*cexp(p->args[1]*t)*(1+p->args[1]*t));
-	return 1;
 }
 
 Boss *stage5_spawn_iku(cmplx pos) {
@@ -150,4 +111,30 @@ Boss *stage5_spawn_iku(cmplx pos) {
 	iku->glowcolor = *RGB(0.2, 0.4, 0.5);
 	iku->shadowcolor = *RGBA_MUL_ALPHA(0.65, 0.2, 0.75, 0.5);
 	return iku;
+}
+
+//TODO: last one to modernize
+int iku_induction_bullet(Projectile *p, int time) {
+	if(time < 0) {
+		return ACTION_ACK;
+	}
+
+	float t = time * sqrt(global.diff);
+
+	if(global.diff > D_Normal && !p->args[2]) {
+		t = time*0.6;
+		t = 230-t;
+		if(t < 0)
+			return ACTION_DESTROY;
+	}
+
+	p->pos = p->pos0 + p->args[0] * t * cexp(p->args[1] * t);
+
+	if(time == 0) {
+		// don't lerp; the spawn position is very different on hard/lunatic and would cause false hits
+		p->prevpos = p->pos;
+	}
+
+	p->angle = carg(p->args[0] * cexp(p->args[1] * t) * (1 + p->args[1] * t));
+	return 1;
 }
--- a/src/stages/stage5/iku.c
+++ b/src/stages/stage5/iku.c
@@ -48,33 +48,26 @@ TASK(iku_slave_visual, { BoxedIkuSlave slave; }) {
 	WAIT(rng_irange(0, period));
 
 	for(;;WAIT(period)) {
-		cmplx offset = (frand()-0.5)*10 + (frand()-0.5)*10.0*I;
-		float alpha = 1;
-
-		Color *clr = RGBA_MUL_ALPHA(0.1*alpha, 0.1*alpha, 0.6*alpha, 0.5*alpha);
-		clr->a = 0;
-
 		PARTICLE(
 			.sprite_ptr = slave->sprites.cloud,
 			.pos = slave->pos,
-			.color = clr,
+			.color = &slave->color,
 			.draw_rule = pdraw_timeout_fade(2, 0.01),
 			.timeout = 50,
+			.move = move_linear(rng_dir()),
 			.flags = PFLAG_REQUIREDPARTICLE,
-			.move = move_linear(offset*0.2),
 		);
 	}
 
 }
 
 void stage5_init_iku_slave(IkuSlave *slave, cmplx pos) {
-	float alpha = 1;
 	slave->pos = pos;
 	slave->spawn_time = global.frames;
 	slave->ent.draw_layer = LAYER_BOSS - 1;
 	slave->ent.draw_func = iku_slave_draw;
 	slave->scale = (1 + I);
-	slave->color = *RGBA_MUL_ALPHA(0.1*alpha, 0.1*alpha, 0.6*alpha, 0.5*alpha);
+	slave->color = *RGBA_MUL_ALPHA(0.1, 0.1, 0.6, 0.5);
 	slave->sprites.lightning0 = res_sprite("part/lightning0");
 	slave->sprites.lightning1 = res_sprite("part/lightning1");
 	slave->sprites.cloud = res_sprite("part/lightningball");
@@ -151,9 +144,9 @@ int iku_induction_bullet(Projectile *p, int time) {
 }
 
 Boss *stage5_spawn_iku(cmplx pos) {
-	Boss *iku = create_boss("Iku Nagae", "iku", pos);
+	Boss *iku = create_boss("Nagae Iku", "iku", pos);
 	boss_set_portrait(iku, "iku", NULL, "normal");
 	iku->glowcolor = *RGB(0.2, 0.4, 0.5);
-    iku->shadowcolor = *RGBA_MUL_ALPHA(0.65, 0.2, 0.75, 0.5);
+	iku->shadowcolor = *RGBA_MUL_ALPHA(0.65, 0.2, 0.75, 0.5);
 	return iku;
 }
--- a/src/stages/stage5/iku.c
+++ b/src/stages/stage5/iku.c
@@ -33,6 +33,7 @@ void iku_lightning_particle(cmplx pos) {
 		.timeout = 20,
 		.draw_rule = pdraw_timeout_fade(0.8, 0.0),
 		.angle = rng_angle(),
+		.flags = PFLAG_MANUALANGLE | PFLAG_NOMOVE | PFLAG_REQUIREDPARTICLE,
 	);
 }
 
--- a/src/stages/stage5/iku.c
+++ b/src/stages/stage5/iku.c
@@ -68,6 +68,7 @@ void stage5_init_iku_slave(IkuSlave *slave, cmplx pos) {
 	slave->sprites.lightning0 = res_sprite("part/lightning0");
 	slave->sprites.lightning1 = res_sprite("part/lightning1");
 	slave->sprites.cloud = res_sprite("part/lightningball");
+	slave->sprites.dot = res_sprite("dot");
 
 	INVOKE_TASK(iku_slave_visual, ENT_BOX(slave));
 }
@@ -113,28 +114,31 @@ Boss *stage5_spawn_iku(cmplx pos) {
 	return iku;
 }
 
-//TODO: last one to modernize
-int iku_induction_bullet(Projectile *p, int time) {
-	if(time < 0) {
-		return ACTION_ACK;
+DEFINE_EXTERN_TASK(iku_induction_bullet) {
+	Projectile *p = TASK_BIND(ARGS.p);
+	cmplx m1 = ARGS.m1;
+	cmplx m2 = ARGS.m2;
+
+	for(int time = 0;; time += WAIT(1)) {
+		int t = time * sqrt(global.diff);
+		if(global.diff > D_Normal && ARGS.mode) {
+			t = time*0.6;
+			t = 230-t;
+		}
+
+		p->pos = p->pos0 + m1 * t * cexp(m2 * t);
+
+		if(time == 0) {
+			// don't lerp; the spawn position is very different on hard/lunatic and would cause false hits
+			p->prevpos = p->pos;
+		}
+
+		p->angle = carg(m1 * cexp(m2 * t) * (1 + m2 * t));
 	}
+}
 
-	float t = time * sqrt(global.diff);
-
-	if(global.diff > D_Normal && !p->args[2]) {
-		t = time*0.6;
-		t = 230-t;
-		if(t < 0)
-			return ACTION_DESTROY;
+DEFINE_EXTERN_TASK(iku_spawn_clouds) {
+	for(int i = 0;; i += WAIT(2)) {
+		iku_nonspell_spawn_cloud();
 	}
-
-	p->pos = p->pos0 + p->args[0] * t * cexp(p->args[1] * t);
-
-	if(time == 0) {
-		// don't lerp; the spawn position is very different on hard/lunatic and would cause false hits
-		p->prevpos = p->pos;
-	}
-
-	p->angle = carg(p->args[0] * cexp(p->args[1] * t) * (1 + p->args[1] * t));
-	return 1;
 }
--- a/src/stages/stage5/iku.c
+++ b/src/stages/stage5/iku.c
@@ -16,6 +16,30 @@
 
 MODERNIZE_THIS_FILE_AND_REMOVE_ME
 
+
+void stage5_init_iku_slave(IkuSlave *slave, cmplx pos) {
+	slave->pos = pos;
+	slave->spawn_time = global.frames;
+
+}
+
+IkuSlave *stage5_midboss_slave(cmplx pos) {
+	IkuSlave *slave = TASK_HOST_ENT(IkuSlave);
+	TASK_HOST_EVENTS(slave->events);
+	stage5_init_iku_slave(slave, pos);
+
+	float alpha = 1;
+	Color *clr = RGBA_MUL_ALPHA(0.1*alpha, 0.1*alpha, 0.6*alpha, 0.5*alpha);
+	clr->a = 0;
+	slave->color = *clr;
+	slave->scale = (1 + I) * 0.7;
+	// TODO: you were here
+	// need to continue porting vvvv that stuff into this functioning
+	// including the PARTICLE
+
+	return slave;
+}
+
 void iku_lightning_particle(cmplx pos, int t) {
 	if(!(t % 5)) {
 		char *part = frand() > 0.5 ? "lightning0" : "lightning1";
@@ -116,3 +140,10 @@ int iku_induction_bullet(Projectile *p, int time) {
 	return 1;
 }
 
+Boss *stage5_spawn_iku(cmplx pos) {
+	Boss *iku = create_boss("Iku Nagae", "iku", pos);
+	boss_set_portrait(iku, "iku", NULL, "normal");
+	iku->glowcolor = *RGB(0.2, 0.4, 0.5);
+    iku->shadowcolor = *RGBA_MUL_ALPHA(0.65, 0.2, 0.75, 0.5);
+	return iku;
+}
--- a/src/stages/stage5/iku.c
+++ b/src/stages/stage5/iku.c
@@ -116,8 +116,8 @@ Boss *stage5_spawn_iku(cmplx pos) {
 
 DEFINE_EXTERN_TASK(iku_induction_bullet) {
 	Projectile *p = TASK_BIND(ARGS.p);
-	cmplx m1 = ARGS.m1;
-	cmplx m2 = ARGS.m2;
+	cmplx radial_vel = ARGS.radial_vel;
+	cmplx angular_vel = ARGS.angular_vel;
 
 	for(int time = 0;; time++, YIELD) {
 		int t = time * sqrt(global.diff);
@@ -126,14 +126,15 @@ DEFINE_EXTERN_TASK(iku_induction_bullet) {
 			t = 230-t;
 		}
 
-		p->pos = p->pos0 + m1 * t * cexp(m2 * t);
+		p->pos = p->pos0 + radial_vel * t * cexp(angular_vel * t);
 
 		if(time == 0) {
 			// don't lerp; the spawn position is very different on hard/lunatic and would cause false hits
 			p->prevpos = p->pos;
 		}
 
-		p->angle = carg(m1 * cexp(m2 * t) * (1 + m2 * t));
+		// calculate angle manually using the derivative of p->pos
+		p->angle = carg(radial_vel * cexp(angular_vel * t) * (1 + angular_vel * t));
 	}
 }
 
--- a/src/stages/stage5/iku.c
+++ b/src/stages/stage5/iku.c
@@ -16,11 +16,53 @@
 
 MODERNIZE_THIS_FILE_AND_REMOVE_ME
 
+static void iku_slave_draw(EntityInterface *e) {
+	IkuSlave *slave = ENT_CAST(e, IkuSlave);
+	//int time = global.frames - slave->spawn_time;
+	r_draw_sprite(&(SpriteParams) {
+		.pos.as_cmplx = slave->pos,
+		.sprite_ptr = slave->sprites.lightning0,
+		.scale.as_cmplx = slave->scale,
+		.color = &slave->color,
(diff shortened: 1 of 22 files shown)


src/stages/stage5/timeline.c
```diff
@@ -400,17 +366,114 @@ static int stage5_superbullet(Enemy *e, int t) {
...
+}) {
+	Enemy *e = TASK_BIND(ARGS.e);
+	cmplx dir = cnormalize(global.plr.pos - e->pos);
+
+	for(int i = -ARGS.count; i <= ARGS.count; i++) {
+		PROJECTILE(
+			.proto = pp_bullet,
+			.pos = e->pos,
+			.color = RGB(0.0, 0.0, 1.0),
+			.move = move_asymptotic_simple(3.5 * cdir(carg(dir) + 0.06 * i), 5),
```
Akaricchi, line 381: It's generally more efficient and cleaner to do `dir * cdir(X)` instead of `cdir(carg(dir) + X)`.

Also in the original code, the bullet speed is 4.5 on Lunatic (that's the `+(global.diff == D_Lunatic)` term). I suggest defining a `real speed = difficulty_value(3.5, 3.5, 3.5, 4.5);` outside the loop

src/stages/stage5/timeline.c
```diff
@@ -400,17 +366,114 @@ static int stage5_superbullet(Enemy *e, int t) {
...
+TASK(greeter_fairy, {
+		cmplx pos;
+		MoveParams move_enter;
+		MoveParams move_exit;
+}) {
+	Enemy *e = TASK_BIND(espawn_fairy_blue_box(ARGS.pos, ITEMS(.points = 1)));
+	e->move = ARGS.move_enter;
+
+	for (int i = 0; i < 5; i++) {
+		INVOKE_SUBTASK_DELAYED(80, greeter_fairy_attack, {
```
Akaricchi, line 396: I think you could rewrite this into a normal function instead of a task. It doesn't have any async control flow, and the delayed invocation can be replaced with just a `WAIT(80)` before the loop. The `WAIT(200)` in the end would have to be shortened to a `WAIT(20)` to match the original behavior.

With delayed invocations it doesn't match the original either by the way, because the loop already waits for `20 * 5 = 100` frames.

Either of those should match:

```c
WAIT(80);
for(int i = 0; i < 5; i++) {
	greeter_fairy_attack(...);
	WAIT(20);
}
WAIT(20);
e->move = ARGS.move_exit;
```

```c
for(int i = 0; i < 5; i++) {
	INVOKE_SUBTASK_DELAYED(80, greeter_fairy_attack, ...);
	WAIT(20);
}
WAIT(100);
e->move = ARGS.move_exit;
```

```c
for(int i = 0; i < 5; i++) {
	INVOKE_SUBTASK_DELAYED(80 + i * 5, greeter_fairy_attack, ...);
}
WAIT(200);
e->move = ARGS.move_exit;
```

I would personally choose number 1 because, as I said, the subtask is not really necessary here. You could also just inline the "attack" function.

src/stages/stage5/timeline.c
```diff
@@ -400,17 +366,114 @@ static int stage5_superbullet(Enemy *e, int t) {
...
+			.count = difficulty_value(1, 2, 3, 4),
+		});
+		WAIT(20);
+	}
+
+	WAIT(200);
+
+	e->move = ARGS.move_exit;
+
+	STALL;
```
Akaricchi, line 407: The `STALL` is not needed in this case. It's generally used to keep subtasks alive, but:
* All subtasks invoked here are short enough that they'll surely be finished before this code is reached;
* If you implement my earlier suggestion, there wouldn't be subtasks at all.

Akaricchi (commented):
Looks mostly fine. Although it's not required to port it perfectly, I left some suggestions on how this could be made a bit more faithful to the original, just to help you sharpen your intuition.

src/stages/stage5/timeline.c
```diff
@@ -334,197 +135,713 @@ static int stage5_magnetto(Enemy *e, int t) {
...
-	AT(2500) {
-		create_enemy1c(VIEWPORT_W/2, 2000, BigFairy, stage5_lightburst, 2.0*I);
+TASK(lightburst_fairies_1, {
+	int num;
+	cmplx pos1;
+	cmplx pos2;
+	cmplx exit;
+}) {
+	for (int i = 0; i < ARGS.num; i++) {
+		cmplx pos = ARGS.pos1 + ARGS.pos2 * i;
```
Akaricchi, line 344: Those parameter names seem misleading. You'd think `pos1` and `pos2` are both absolute positions, but in fact `pos2` seems to be used like an offset to `pos1`.

src/stages/stage5/timeline.c
```diff
@@ -334,197 +135,713 @@ static int stage5_magnetto(Enemy *e, int t) {
...
+		WAIT(40);
 	}
+}
 
-	AT(2900)
-		global.boss = stage5_spawn_midboss();
+TASK(loop_swirl_move, {
+    MoveParams *move;
+	cmplx turn;
+	cmplx retention;
```
Akaricchi, line 374: Bad indent

src/stages/stage5/timeline.c
```diff
@@ -334,197 +135,713 @@ static int stage5_magnetto(Enemy *e, int t) {
...
+}) {
+	int begin = creal(ARGS.turn) + 1;
+	int end = cimag(ARGS.turn) - 1;
+	int time = end - begin;
+
+	ARGS.move->retention = ARGS.retention;
+	WAIT(time / 1.5);
```
Akaricchi, line 381: Why `/ 1.5`?

src/stages/stage5/timeline.c
```diff
@@ -334,197 +135,713 @@ static int stage5_magnetto(Enemy *e, int t) {
...
-		// TODO: determine if this hack is still required with the new dialogue system
-		global.timer++;
+    e->move = move_linear(ARGS.velocity);
+    INVOKE_SUBTASK_DELAYED(80, loop_swirl_move,
+        .move = &e->move,
+		.turn = ARGS.turn,
+		.retention = ARGS.retention
+    );
+
+	int difficulty = 30 - difficulty_value(4, 8, 16, 24);
```
Akaricchi, line 400: Broken indents

src/stages/stage5/timeline.c
```diff
@@ -334,197 +135,713 @@ static int stage5_magnetto(Enemy *e, int t) {
...
+		for(int i = 1; i >= -1; i -= 2) {
+			PROJECTILE(
+				.proto = pp_bullet,
+				.pos = e->pos,
+				.color = RGB(0.3, 0.4, 0.5),
+				.move = move_asymptotic_simple(i * 2 * ARGS.velocity * I / cabs(ARGS.velocity), 3),
+				// TODO: bullets only fire down-left with the above
+				// they should also fire in other directions correlating to the angle of "turn" the swirls make
+				//.rule = asymptotic,
+				//.args = { i*2*e->args[0]*I/cabs(e->args[0]), 3 }
```
Akaricchi, line 412: Sounds like you want to use `e->move.velocity` instead of `ARGS.velocity` then.

Also, you have a `velocity / cabs(velocity)` there, which should be simplified to `cnormalize(velocity)`

src/stages/stage5/timeline.c
```diff
@@ -334,197 +135,713 @@ static int stage5_magnetto(Enemy *e, int t) {
...
+
+	int difficulty = difficulty_value(0.25, 0.50, 0.75, 1);
+	for (int x = 0; x < 400; x++) {
+		double base_angle = carg(global.plr.pos - e->pos);
 
-	FROM_TO(3300, 4000, 90-10*global.diff)
-		create_enemy1c(200.0*I+VIEWPORT_W*(_i&1), 1500, Fairy, stage5_superbullet, 3-6*(_i&1));
+		for (int i = 1; i >= -1; i -= 2) {
+			double a = i * 0.2 - 0.1 * difficulty + i * 3.0 / (x + 1);
+			cmplx aim = cdir(base_angle + a);
```
Akaricchi, line 453: 1. Don't use `double`, use `real`
2. Better store the aiming direction vector e.g. `cmplx plraim = cnormalize(global.plr.pos - e->pos)`. Then you can rotate it by an arbitrary angle with `cmplx aim = plraim * cdir(a)`

src/stages/stage5/timeline.c
```diff
@@ -334,197 +135,713 @@ static int stage5_magnetto(Enemy *e, int t) {
...
-		create_enemy1c(VIEWPORT_W/2, 2000, BigFairy, stage5_lightburst, 2.0*I);
+	int difficulty = difficulty_value(8, 6, 4, 3);
+
+	for (int x = 0; x < 600; x++) {
+		PROJECTILE(
+			.proto = pp_rice,
+			.pos = e->pos + 20 * cdir(2.0 * M_PI * rng_real()),
+			.color = RGB(0, 0, 255),
+			.move = move_linear(cdir(2.0 * M_PI * rng_real())),
+		);
```
Akaricchi, line 501: Do not put multiple RNG calls in the same statement. The evaluation order is undefined and may break replays across builds (because C is awesome like that). We had the `a*rand*` facility (and its successor, `RNG_ARRAY` and `vrng_*`) to hack around this problem, but it's usually better to just use variables.

`ninja upkeep` can detect problematic RNG usage, although it has false positives.

Note that this is also sometimes neglected in effect-spawning code, because it ultimately has no effect on game state, but that's still not a good practice.

Akaricchi:
Github was "helpful" enough to hide some of my comments, beware.

src/stages/stage5/timeline.c
StarWitch: It's the only way to get the swirls to end their "looping" motion at the correct angle. I'm not too sure why, as I was mostly roughing it out at the time.

Akaricchi: That's not the right way to go about fixing it. You should either change the rotation angle-per-frame (the argument to `cdir()` used for `retention`), or adjust the begin/end timing at call site. There are also a couple more problems with this code that I haven't noticed at first:

* First of all there's no reason to express the timings as a single `cmplx` parameter. The only reason the old code does it is because in the crappy old system you could only have `cmplx` args (and only 4 of them at that). You should just make `int end, begin` task parameters; I'd name them `turn_begin/turn_end` for clarity.
* The +/-1 offsets should be just moved to the call site.
* The `begin` timing isn't used properly. You should `WAIT(begin)` before changing retention to a rotating one. And then you shouldn't use `INVOKE_SUBTASK_DELAYED` - set the `begin` timing properly instead.
* Does the `WAIT()` even work? It seems like you're passing a negative number into it:
    * You are calling the `loop_swirls` task with `turn = 9`. This is looks very wrong...
    * `begin == creal(9) + 1 == 9 + 1 == 10`
    * `end == cimag(9) - 1 == 0 - 1 == -1`
    * `time == end - begin == -1 - 10 == -11`
    * `WAIT(time / 1.5) = WAIT(-7)` - honestly not even sure what happens in this case. We probably should have an assertion for wait time >= 0 inside that macro.


Once you get this all figured out (and preferably *not* before you do), I'd recommend switching to a more intuitive set of outermost parameters for turning:
* `int turn_delay` - time to wait before initiating the turn, in frames.
* `int turn_duration` - duration of the turn, in frames.
* `real turn_angle` - total angle of the turn, in radians.

This is exactly what the stage 2 fairy turning code does, and perhaps it makes sense to generalize and move it to `common_tasks.c` eventually.

StarWitch: Done and done. I was worried the movement was going to be to different but wiping out the old format and using what was in Stage 2 ended up being much better. I also moved it to `common_tasks`. I did a before/after comparison and the movement in Stage 2 seemed identical, so there you go.

src/stages/stage5/timeline.c
```diff
@@ -37,537 +27,854 @@ static void stage5_dialog_post_midboss(void) {
...
-	}
+	e->move = move_towards(e->pos + ARGS.acceleration * 70 + ARGS.offset, 0.05);
+	WAIT(60);
 
-	AT(6960) {
-		stage_unlock_bgm("stage5");
-		global.boss = stage5_spawn_boss();
+	int difficulty = difficulty_value(1, 2, 3, 4);
+	for (int x = 0; x < 140; x++) {
+		cmplx n = cdir(M_PI * sin(x / (8.0 + difficulty) + rng_real() * 0.1) * carg(global.plr.pos - e->pos));
```
Akaricchi, line 594: Looks suspiciously wrong. The original line was this, I think?
```c
cmplx n = cexp(I*M_PI*sin(_i/(8.0+global.diff)+frand()*0.1)+I*carg(global.plr.pos-e->pos));
```

In that case it is almost correct, except instead of multiplying by the `carg(...)` you have to add it, like the original does... and then you preferably get rid of the `carg`. I explain that part in detail below.

The expression is confusing so let's break it down.

**Step 0**: de-laochainlanize the code:
```c
cmplx n = cexp(I * M_PI * sin(_i / (8.0 + global.diff) + frand() * 0.1) + I * carg(global.plr.pos - e->pos));
```

**Step 1**: replace `cexp(I * x)` with `cdir(x)`. There are 2 `I`s here, so let's factor them out:
```c
cmplx n = cexp(I * (M_PI * sin(_i / (8.0 + global.diff) + frand() * 0.1) + carg(global.plr.pos - e->pos)));
```
Then we can substitute `cdir()` easily:
```c
cmplx n = cdir(M_PI * sin(_i / (8.0 + global.diff) + frand() * 0.1) + carg(global.plr.pos - e->pos));
```

**Step 2**: get rid of that `carg`:

What we have here is a sum of two angles, `A` and `B`
```c
real A = M_PI * sin(_i / (8.0 + global.diff) + frand() * 0.1);
real B = carg(global.plr.pos - e->pos);
cmplx n = cdir(A + B);
```

Since complex multiplication is equivalent to a scale and a rotation, we can rewrite `n` as follows:
```c
cmplx n = cdir(A) * cdir(B);
```

If we represent A and B as direction vectors rather than angles, we get:
```c
cmplx A = cdir(M_PI * sin(_i / (8.0 + global.diff) + frand() * 0.1));
cmplx B = cdir(carg(global.plr.pos - e->pos));
cmplx n = A * B;
```

Let's focus on B now:
```c
cdir(carg(global.plr.pos - e->pos))
```
This expression basically says "get the angle (i.e. argument) of complex `global.plr.pos - e->pos`, and make another complex with the same angle but magnitude = 1". That's a very roundabout way of spelling `cnormalize(global.plr.pos - e->pos)`.
It's also mathematically equivalent to `(global.plr.pos - e->pos) / cabs(global.plr.pos - e->pos)` (`cnormalize` does a slightly optimized version of that).

So:
```c
cmplx A = cdir(M_PI * sin(_i / (8.0 + global.diff) + frand() * 0.1));
cmplx B = cnormalize(global.plr.pos - e->pos);
cmplx n = A * B;
```

Folding A and B back into n:
```c
cmplx n = cdir(M_PI * sin(_i / (8.0 + global.diff) + frand() * 0.1)) * cnormalize(global.plr.pos - e->pos);
```

### TL;DR

To deal with such expressions you just need to know the following identities:

```
cdir(A + B)  ==  cdir(A) * cdir(B)
```
```
cdir(carg(X))  ==  cnormalize(X)
```
And from these follows:
```
cdir(A + carg(B))  ==  cdir(A) * cnormalize(B)
```

Akaricchi (commented):
Not a super comprehensive review, admittedly I skimmed through some parts so apologies if I missed anything important; I'll come back to it later. Also github is probably going to hide some of my comments again.

StarWitch:
One thing to note is that some of the `timeline.c` enemy timings need to be tweaked a bit. I tried to replicate them 1:1 with the old system but some appear too soon/too late for various reasons. I'll work on that once `Overload` is finished.

src/stages/stage5/spells/artificial_lightning.c
```diff
@@ -10,148 +10,155 @@
...
+	Enemy *e = TASK_BIND(create_enemy1c(ARGS.pos, ENEMY_IMMUNE, NULL, NULL, 0));
 
-	FROM_TO(0,200,20)
-		e->args[0] *= cexp(I * (0.25 + 0.25 * frand() * M_PI));
+	INVOKE_TASK(lightning_slave_move, ENT_BOX(e), ARGS.move_arg);
 
-	FROM_TO(0, 200, 3)
-		if(cabs(e->pos-global.plr.pos) > 60) {
-			Color *clr = RGBA(1-1/(1+0.01*_i), 0.5-0.01*_i, 1, 0);
+	for(int x = 0; x < 67; x += WAIT(3)) {
```
laochailan, line 87: what is `x += WAIT(3)`? i am not a fan of these strange for-loop reinvention constructions. but if it has to be done, `x++, WAIT(3)` is better

src/stages/stage5/spells/artificial_lightning.c
```diff
@@ -10,148 +10,155 @@
...
+				// TODO: this doesn't quite work right
+				.move = move_asymptotic_simple(0.75 * ARGS.move_arg / cabs(ARGS.move_arg) * I, 10),
 			);
 
 			if(projectile_in_viewport(p)) {
 				for(int i = 0; i < 3; ++i) {
-					tsrand_fill(2);
-					iku_lightning_particle(p->pos + 5 * afrand(0) * cexp(I*M_PI*2*afrand(1)), 0);
+					RNG_ARRAY(rand, 2);
+					iku_lightning_particle(p->pos + 5 * vrng_sreal(rand[0]) * cdir(M_TAU * vrng_sreal(rand[1])));
```
laochailan, line 102: `vrng_sreal` -> `vrng_real` although here it doesn’t affect the result.

src/stages/stage5/spells/atmospheric_discharge.c
```diff
@@ -10,61 +10,61 @@
...
-		cmplx p2 = p1 + (120+20*global.diff)*cexp(0.5*I-afrand(2)*I)*(1-2*(afrand(3) > 0.5));
-
-		int i;
-		int c = 6+global.diff;
+TASK(ball_shoot, NO_ARGS) {
+	int difficulty = difficulty_value(1, 2, 3, 4);
+	for(int x = 0;; x += WAIT(difficulty_value(19, 17, 15, 13))) {
+		RNG_ARRAY(rand, 4);
+		cmplx p1 = VIEWPORT_W * vrng_real(rand[0]) + VIEWPORT_H/2 * I * vrng_real(rand[1]);
+		cmplx p2 = p1 + (120 + 20 * difficulty) * cdir(0.5 - vrng_real(rand[2])) * (1 - 2 * (vrng_real(rand[3]) > 0.5));
```
laochailan, line 38: `+ difficulty_value(140, 160, 180, 200)` or `+20*difficulty_value(7,8,9,10)`

`cdir(0.5 - vrng_real(rand[2]))` -> `cdir(0.5*vrng_sreal(rand[2]))`

for the last one maybe `vrng_chance(rand[3], 0.5)` is better?

src/stages/stage5/spells/atmospheric_discharge.c
```diff
@@ -10,61 +10,61 @@
...
 
-		for(i = -c*0.5; i <= c*0.5; i++) {
+		int count = difficulty_value(7, 8, 9, 10);
```
laochailan, line 40: this should be 1,2,3,4 as it replaces global.diff below!

src/stages/stage5/spells/atmospheric_discharge.c
```diff
@@ -10,61 +10,61 @@
...
 				.proto = pp_ball,
-				.pos = p1+(p2-p1)/c*i,
-				.color = RGBA(1-1/(1+fabs(0.1*i)), 0.5-0.1*abs(i), 1, 0),
-				.rule = accelerated,
-				.args = {
-					0, (0.004+0.001*global.diff)*cexp(I*carg(p2-p1)+I*M_PI/2+0.2*I*i)
-				},
+				.pos = p1 + (p2 - p1) / count * i,
+				.color = RGBA(1 - 1 / (1 + fabs(0.1 * i)), 0.5 - 0.1 * abs(i), 1, 0),
+				.move = move_accelerated(0, (0.004 + 0.001 * difficulty) * cnormalize(p2 - p1) * cdir(M_PI/2 + 0.2 * i)),
```
laochailan, line 46: better yet:
`0.001*difficulty_value(5,6,7,8)`

src/stages/stage5/timeline.c
```diff
@@ -37,537 +26,845 @@ static void stage5_dialog_post_midboss(void) {
...
-			cmplx n = cexp(I*carg(global.plr.pos) + 2.0*I*M_PI/c*i);
+TASK(magnetto_swirl_shoot, {
+	BoxedEnemy e;
+	cmplx move_to;
+}) {
+	Enemy *e = TASK_BIND(ARGS.e);
+
+	int interval = difficulty_value(12, 9, 6, 3);
+	int bullets = difficulty_value(1, 2, 2, 2);
+	int bullet_dir = difficulty_value(0, 0, 1, 0);
```
laochailan, line 39: why is a direction valued 0 and 1?

src/stages/stage5/timeline.c
```diff
@@ -37,537 +26,845 @@ static void stage5_dialog_post_midboss(void) {
...
-				.pos = e->pos + 50*n*cexp(-1.0*I*_i*global.diff),
-				.color = RGB(0.3, 0, 0.7+0.3*(_i&1)),
-				.rule = asymptotic,
-				.args = {
-					2.5*n+0.25*global.diff*afrand(0)*cexp(2.0*I*M_PI*afrand(1)),
-					3
-				}
+				.pos = e->pos + 50 * n * cdir(-1.0 * x * difficulty),
+				.color = RGB(0.3, 0, 0.7 + 0.3 * (i % 2)),
+				.move = move_asymptotic_simple(2.5 * n + 0.25 * difficulty * rng_sreal() * rng_dir(), 3),
```
laochailan, line 278: RNG POLICE, OPEN UP!!!

src/stages/stage5/timeline.c
```diff
@@ -37,537 +26,845 @@ static void stage5_dialog_post_midboss(void) {
...
-		create_enemy1c(VIEWPORT_W*(_i&1)+70.0*I+50*_i*I, 400, Fairy, stage5_greeter, 3-6*(_i&1));
-	}
+	int delay = difficulty_value(6, 5, 4, 3) + ARGS.reduction;
+	int amount = 700 / delay;
+	int difficulty = difficulty_value(1, 2, 3, 4);
 
-	FROM_TO(270, 320, 40)
-		create_enemy1c(VIEWPORT_W/4+VIEWPORT_W/2*_i, 2000, BigFairy, stage5_lightburst, 2.0*I);
+	for(int x = 0; x < amount; x++) {
+		cmplx n = cnormalize(global.plr.pos - e->pos) * cdir((0.2 - 0.02 * difficulty) * x);
```
laochailan, line 368: `0.2 - 0.02 * difficulty`
-> `0.02 * difficulty_value(9,8,7,6)`

src/stages/stage5/timeline.c
```diff
@@ -37,537 +26,845 @@ static void stage5_dialog_post_midboss(void) {
...
+	e->move = move_linear(ARGS.velocity);
 
-	AT(5360) {
-		create_enemy1c(30.0*I+VIEWPORT_W, 1500, Fairy, stage5_superbullet, -2 + I);
+	int difficulty = difficulty_value(8, 6, 4, 3);
 
+	for(int x = 0; x < 600; x++) {
+		PROJECTILE(
+			.proto = pp_rice,
+			.pos = e->pos + 20 * cdir(M_TAU * rng_real()),
```
laochailan, line 519: rng_dir

src/stages/stage5/timeline.c
```diff
@@ -37,537 +26,845 @@ static void stage5_dialog_post_midboss(void) {
...
+			.color = RGB(0, 0, 255),
+			.move = move_linear(cdir(M_TAU  * rng_real())),
```
laochailan, line 521: rng_dir

laochailan (commented):
I’m not sure about replacing every `global.diff` by `difficulty = difficulty_value(1,2,3,4)` if no further simplifications are possible. I tried to do the simplifications where it came up in my comments.

And I probably missed stuff some task-related subtleties!

src/stages/stage5/spells/artificial_lightning.c
```diff
@@ -10,148 +10,155 @@
...
+	for(int time = 0;; time++, YIELD) {
+		if(time > 0 && !(time % 100)) {
+			int count = difficulty_value(7, 7, 7, 9);
+			for(int i = 0; i < count; i++) {
+				Projectile *p = PROJECTILE(
+					.proto = pp_bigball,
+					.pos = boss->pos,
+					.color = RGBA(0.5, 0.1, 1.0, 0.0),
+				);
+				INVOKE_SUBTASK(zigzag_move, ENT_BOX(p), cdir(M_TAU / count * i) * cnormalize(global.plr.pos - boss->pos));
```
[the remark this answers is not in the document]
Akaricchi: No it shouldn't be, it already binds to the projectile lifetime which should be enough.

src/stages/stage5/spells/artificial_lightning.c
```diff
@@ -10,148 +10,155 @@
...
-static int lightning_slave(Enemy *e, int t) {
-	if(t < 0)
-		return 1;
-	if(t > 200)
-		return ACTION_DESTROY;
+TASK(zigzag_shoot, { BoxedBoss boss; } ) {
+	Boss *boss = TASK_BIND(ARGS.boss);
+	for(int time = 0;; time++, YIELD) {
+		if(time > 0 && !(time % 100)) {
+			int count = difficulty_value(7, 7, 7, 9);
```
Akaricchi, line 54: I think you want something like this instead

```c
for(int time = 0;;) {
	time += WAIT(delay);
	int count = difficulty_value(7, 7, 7, 9);
```

src/stages/stage5/spells/artificial_lightning.c
Akaricchi: @laochailan that's not equivalent. `WAIT()` returns the amount of frames waited, so you'd have to do `x += 3, WAIT(3)` if you really wanted to rewrite this. But that's repeating yourself, and not any better in my opinion.

src/stages/stage5/spells/artificial_lightning.c
Akaricchi: Actually just replace that whole `cdir(M_TAU * vrng_sreal(rand[1])))` with `vrng_dir(rand[1])`.

src/stages/stage5/spells/atmospheric_discharge.c
Akaricchi: Doesn't seem to be the case? `int difficulty = difficulty_value(1, 2, 3, 4);` at the top of the task replaces `global.diff` (not very useful in my opinion, but correct). The `count` here replaces `int c = 6+global.diff;`, which seems fine. Am I missing something?

src/stages/stage5/spells/atmospheric_discharge.c
Akaricchi: Also in general I would prefer to have all the constant `difficulty_value(...)` expressions grouped up in variables outside of all the loops, preferably with meaningful names, for easier tuning and general readability.

src/stages/stage5/spells/artificial_lightning.c
laochailan: then this is wrong because the amount of iterations has been shortened to 200/3.

src/stages/stage5/spells/atmospheric_discharge.c
laochailan: no i misread

src/stages/stage5/spells/artificial_lightning.c
Akaricchi: Ah yes, you are correct. I missed the context.

The `_i` that the `x` is replacing is the iteration count (i think), not a timer. So it makes no sense to add the wait time to it. It should indeed be `x++, WAIT(3)`, or more "conventionally":
```c
for(int x = 0; x < 67; x++) {
	/* the loop... */
	WAIT(3);
}
```

src/stages/stage5/iku.c
```diff
@@ -14,105 +14,131 @@
...
+	Projectile *p = TASK_BIND(ARGS.p);
+	cmplx m1 = ARGS.m1;
+	cmplx m2 = ARGS.m2;
 
-	float t = time*sqrt(global.diff);
+	for(int time = 0;; time++, YIELD) {
+		int t = time * sqrt(global.diff);
+		if(global.diff > D_Normal && ARGS.mode) {
+			t = time*0.6;
+			t = 230-t;
```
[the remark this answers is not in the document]
StarWitch: What would that look like? Considering `time` is always ticking upwards, when would `time < 0` ever happen?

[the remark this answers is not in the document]
laochailan: if `time*0.6 > 230`, `t < 0` is true. In the original rule this was some kind of time out. Now you could use INVOKE_SUBTASK_DELAYED with the common_enemy_kill (or what it was called) task to do it.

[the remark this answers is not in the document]
laochailan: i mean the projectile equivalent because it is a projectile

src/stages/stage5/timeline.c
StarWitch: There's a few places in the original code where it specifically checks against `D_Hard` and `D_Lunatic` being defined. Example:

```
			PROJECTILE(
				.proto = pp_ball,
				.pos = e->pos,
				.color = RGBA(0.1 + 0.5 * pow((t - 140) / 140.0, 2), 0.0, 0.8, 0.0),
				.rule = accelerated,
				.args = {
					(-2 + (global.diff == D_Hard)) * dir, // <--- this line
					0.02 * dir * (!i || global.diff != D_Lunatic), // <--- also this line
				},
```

It seems a bit silly to me, but I was trying to replicate original behaviour as much as possible.

StarWitch:
Fixed the merge conflict, and rebased it against `remastered-staging`.

src/common_tasks.c
```diff
@@ -241,6 +241,19 @@ int common_charge_custom(
...
+	int delay,
+	int duration,
+	real angle
+) {
+	WAIT(delay);
+	cmplx r = cdir(angle / duration);
+	move->retention *= r;
+	WAIT(duration);
+	move->retention /= r;
+}
```
laochailan, line 255: `delay` is not needed because `INVOKE_TASK_DELAYED` exists

src/stages/stage5/iku.c
```diff
@@ -14,105 +14,132 @@
...
+
+void iku_lightning_particle(cmplx pos) {
+	char *part = rng_chance(0.5) ? "lightning0" : "lightning1";
+	PARTICLE(
+		.sprite = part,
+		.pos = pos,
+		.color = RGBA(1.0, 1.0, 1.0, 0.0),
+		.timeout = 20,
+		.draw_rule = pdraw_timeout_fade(0.8, 0.0),
+		.angle = rng_real() * M_TAU,
```
laochailan, line 35: `rng_angle`

src/stages/stage5/nonspells/boss_nonspell_2.c
```diff
@@ -10,55 +10,58 @@
 
 #include "nonspells.h"
 
-MODERNIZE_THIS_FILE_AND_REMOVE_ME
-
 static cmplx bolts2_laser(Laser *l, float t) {
 	if(t == EVENT_BIRTH) {
+		l->shader = res_shader_optional("lasers/iku_lightning");
```
laochailan, line 15: Lasers shaders are obsolete now right? This causes
```
  W: _get_resource: shader program 'lasers/iku_lightning' was not preloaded
102948    D: vfs_set_error: Node 'res/shader/lasers/iku_lightning.prog' does not exist
102948    E: load_shader_program_stage1: VFS error: Node 'res/shader/lasers/iku_lightning.prog' does not exist
```

laochailan (changes requested):
I tried to play the stage but it crashed when the `iku_lightning` shader was loaded and not found.

src/stages/stage5/nonspells/boss_nonspell_2.c
StarWitch: Fixed!

laochailan (approved):
looking good now. ready for merging?

StarWitch:
Only thing holding back this PR is the `overload` spell. It's currently using the old system because of how galaxy brain it is. It still functions in-situ, of course, and maybe now that we have live-reloading work on it could progress faster.

My own idea is to just get this (and the other stage PRs) merged if the bulk of the coroutine/modernization work is done, so we can start fine-tuning and ironing out wrinkles (like `overload`) instead of feeling like these are huge PRs we have to completely context-change in order to understand fully. I feel like it'd actually get us a v1.4 release this year.

laochailan:
I agree. If it works in situ it should be fine.

Akaricchi:
Merged in 46ad7b183bd5ed97afa45d637ab4b6e2647449f4

PR #286, opened 2021-04-26
Windows test builds, cleanup of CI

StarWitch:
Adds a Windows test build, cleans up some of the CI flow.

Related: #278


StarWitch:
Only thing that's failing is that `timespec_get` issue, as before. Seems to be an issue with building against anything before 10.15.

.github/workflows/main.yml
```diff
@@ -27,82 +27,120 @@ jobs:
...
       env:
         TAISEI_NOPRELOAD: 0
         TAISEI_PRELOAD_REQUIRED: 1
 
     - name: Upload Log
       uses: actions/upload-artifact@v2
-      if: failure()
       with:
-        name: taisei_linux_fail_log
+        name: taisei_linux_release.log
```
Akaricchi, line 65: Why is it called "release" now? That's a bit confusing since this doesn't make release builds.

scripts/ci/linux-x86_64-build-test-ci.ini
```diff
@@ -0,0 +1,22 @@
...
+ldflags = []
+
+[built-in options]
+c_args = cflags
+c_link_args = ldflags
+cpp_args = cflags
+cpp_link_args = ldflags
+wrap_mode = 'nofallback'
+force_fallback_for = 'cglm'
+deprecation_warnings = 'no-error'
```
Akaricchi, line 15: `deprecation_warnings` is a taisei project option, not a builtin. It should go into `[project options]`, otherwise I don't think it'll work. Also, it seems to be missing from the macos config.

scripts/ci/windows-llvm_mingw-x86_64-build-test-ci.ini
```diff
@@ -0,0 +1,66 @@
...
+windres = exeprefix+'windres'
+ld = exeprefix+'ld'
+
+[built-in options]
+c_args = cflags
+c_link_args = ldflags
+cpp_args = cflags
+cpp_link_args = ldflags
+wrap_mode = 'forcefallback'
+deprecation_warnings = 'no-error'
```
Akaricchi, line 43: ditto

.github/workflows/main.yml
```diff
@@ -27,82 +27,120 @@ jobs:
...
-      run: meson compile -C build/
-
-    - name: Install - Step 3 (Meson)
-      run: meson install -C build/
-
-    - name: Run - Step 4 (Taisei)
-      run: $(pwd)/build-test/bin/taisei -R $(pwd)/misc/tests/test-replay.tsr
+      run: |
+        sudo apt update
+        sudo apt install -y -qq build-essential libsdl2-dev libogg-dev libopusfile-dev libpng-dev libzip-dev libx11-dev libwayland-dev python3-docutils libwebp-dev libfreetype6-dev libzstd-dev
```
Akaricchi, line 41: Probably makes sense to break this into one package per line; it's more readable and git-friendly. Same with brew on macos and the other apt command on the windows builder.

Also, we probably don't need sudo here.

StarWitch: Unfortunately, `sudo` is required, because Ubuntu.

.github/workflows/main.yml
```diff
@@ -27,82 +27,159 @@ jobs:
...
+        libwayland-dev \
+        python3-docutils \
+        libwebp-dev \
+        libfreetype6-dev \
+        libzstd-dev
+        # pip
+        pip install \
+        meson==0.55.3 \
+        ninja \
+        zstandard
```
Akaricchi, line 58: There's actually a cleaner YAML syntax for this that doesn't require the backslashes:

```suggestion
      run: >
        sudo apt update

        sudo apt install -y -qq
          build-essential
          libsdl2-dev
          libogg-dev
          libopusfile-dev
          libpng-dev
          libzip-dev
          libx11-dev
          libwayland-dev
          python3-docutils
          libwebp-dev
          libfreetype6-dev
          libzstd-dev

        pip install
          meson==0.55.3
          ninja
          zstandard
```

Note the extra blank lines between commands.

.github/workflows/main.yml
```diff
@@ -27,82 +27,159 @@ jobs:
...
+
+    # TODO: turn -Dwerror back on when build is fixed
+    - name: Configure (Meson)
+      run: |
+        meson setup build/ \
+          --prefix=$(pwd)/build-test \
+          --native-file scripts/ci/linux-x86_64-build-test-ci.ini
```
Akaricchi, line 65: I'm not sure `scripts/` is the right place for this. How about putting both the machine files and the test replay into `misc/ci`, along with any further CI-only files we might need in the future?

.github/workflows/main.yml
StarWitch: Sure. Do you want me to do that with the rest of the CI tests as well?

Akaricchi: Yeah that would be good.

PR #288, opened 2021-04-28
Use 'create-dmg' on macOS for prettier .dmg files

StarWitch:
Like it says on the tin.

This would even let us use background images in the future, should we get art for that (or care enough to do it).

Still lets you use `genisoimage` on non-macOS platforms for testing purposes.

### DIFF
--- a/scripts/macos-gen-dmg.py
+++ b/scripts/macos-gen-dmg.py
@@ -32,24 +32,21 @@ def main(args):
         # for compiling on Darwin as 'genisoimage' doesn't exist
         # uses 'create-dmg' (brew install create-dmg)
         if platform == "darwin":
-            command = shlex.split(' \
-                create-dmg \
-                --volname "Taisei" \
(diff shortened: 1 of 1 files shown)


scripts/macos-gen-dmg.py
```diff
@@ -28,12 +29,34 @@ def main(args):
...
+                --window-size 550 480 \
+                --icon-size 64 \
+                --icon "Taisei.app" 100 50 \
+                --icon "README.txt" 50 200 \
+                --icon "STORY.txt" 200 200 \
+                --icon "GAME.html" 350 200 \
+                --icon "COPYING" 125 350 \
+                --icon "ENVIRON.html" 275 350 \
+                --hide-extension "Taisei.app" \
+                --app-drop-link 300 50') + [
```
Akaricchi, line 49: Maybe use a triple-quoted string instead of escaping linebreaks like that? Or just use a plain argument list instead of `shlex.split`ting it; will probably be only a little more verbose.

PR #289, opened 2021-04-30
Release Automation

StarWitch:
A lot of trial-and-error went into this, but here it is.

General steps in this:

1. Create a `v*.*.*[-alpha/-rc/-etc]` **branch**, wait for the `Release Builds` job to finish, download the artifacts on the job page in `Actions` (near the bottom of the page) (or use the API), as well as the build logs, verify them.
2. If everything looks good, push to the tag (i.e `refs/tags/v*.*.*[-alpha/-rc/-etc]`).
3. It will build the files AGAIN, and you can double-check them there too.
4. Upload them as usual to the official tag release page. (At least until the Publishing pipeline is also reviewed/merged, split off from this one.)

Most of what's present is already done in the CI pipeline, but using different machine files to release a more optimized build, as well as extra steps for packaging. There's a little bit of duplication between CI and Release, but it's easier to read and maintain than trying to write a bunch of conditionals inside a single `.yaml` file to do both CI and release builds.

I've also made it so it produces `.sig` and `.sha256` files for each one of the releases, and added some Python scripting to help that along.

Here's what it looks like in action: https://github.com/StarWitch/taisei/releases/tag/v1.3.50 (ignore the `v1.3.50`, it's just for testing)

### Post-Merge Steps

1. Adding a `GPG_PRIVATE_KEY` and `GPG_PASSPHRASE` to `Settings -> Secrets` inside the main Taisei project so the pipeline can generate `.sig` files. These files are hidden in normal pipeline use and I've made sure to fork our version of `ghaction-import-gpg` at [taisei-project/ghaction-import-gpg](https://github.com/taisei-project/ghaction-import-gpg).


.github/workflows/release.yml
```diff
@@ -0,0 +1,606 @@
...
+        zstandard
+
+    # install 3.x on top of older Ubuntu manually as PPA does not exist
+    - name: Install NSIS 3.x
+      run: >
+        wget http://security.ubuntu.com/ubuntu/pool/universe/n/nsis/nsis-pluginapi_3.05-2_all.deb -P /tmp
+
+        wget http://security.ubuntu.com/ubuntu/pool/universe/n/nsis/nsis-common_3.05-2_all.deb -P /tmp
+
+        wget http://security.ubuntu.com/ubuntu/pool/universe/n/nsis/nsis_3.05-2_amd64.deb -P /tmp
```
Akaricchi, line 264: I think you can combine this into one command.

.github/workflows/release.yml
```diff
@@ -0,0 +1,606 @@
...
+      id: vars
+      run: |
+        git fetch --force --tags
+        echo ::set-output name=tag::${GITHUB_REF#refs/tags/v}
+
+    - name: Compress Source
+      run: >
+        tar --exclude-vcs -cf - ./ |
+        xz -9e --threads=8 -c - >
+        Taisei-${{ steps.vars.outputs.tag }}-source.tar.xz
```
Akaricchi, line 598: We rely on `meson dist` to package sources. It does a bit more than this, including running custom dist scripts registered from the build system (which we currently do not use, but we might eventually). Unfortunately it requires a working build directory, and it also does a test build with a clean source dir and default options as a sanity check (this can be disabled with `--no-tests`). For practical purposes, we can probably run it as a step in the Linux builder instead of its own container.

The naming scheme of the source archives is also different — `taisei-v{{version}}.tar.xz`. Unfortunately inconsistent with binary packages, but I'd prefer not to change it for now, to minimize friction for package maintainers and whoever else might rely on our URLs.

Akaricchi (commented):
Not a thorough review, just addressing what sticks out the most for now.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,179 @@
...
+            echo "taisei_changelog_tag=default" >> $GITHUB_ENV
+          else
+            echo "taisei_changelog_tag=${{ github.event.inputs.tag }}" >> $GITHUB_ENV
+          fi
+
+      - name: Import GPG Key
+        uses: crazy-max/ghaction-import-gpg@v1
+        env:
+           GPG_PRIVATE_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
+           PASSPHRASE: ${{ secrets.GPG_PASSPHRASE }}
```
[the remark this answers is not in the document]
StarWitch: For this, we could easily just have the `gpg` command read from `stdin` to import the key from the GitHub Secrets.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,179 @@
...
+          --detach-sig /opt/windows-x86-${{ env.taisei_version }}-setup-exe/Taisei-${{ env.taisei_version }}-setup-x86.exe
+
+          gpg --output /opt/emscripten-wasm32-${{ env.taisei_version }}-txz/Taisei-${{ env.taisei_version }}-emscripten-wasm32.tar.xz.sig
+          --detach-sig /opt/emscripten-wasm32-${{ env.taisei_version }}-txz/Taisei-${{ env.taisei_version }}-emscripten-wasm32.tar.xz
+
+          gpg --output /opt/switch-aarch64-${{ env.taisei_version }}-zip/Taisei-${{ env.taisei_version }}-switch-aarch64.zip.sig
+          --detach-sig /opt/switch-aarch64-${{ env.taisei_version }}-zip/Taisei-${{ env.taisei_version }}-switch-aarch64.zip
+
+          gpg --output /opt/source-${{ env.taisei_version }}-txz/Taisei-${{ env.taisei_version }}-source.tar.xz.sig
+          --detach-sig /opt/source-${{ env.taisei_version }}-txz/Taisei-${{ env.taisei_version }}-source.tar.xz
```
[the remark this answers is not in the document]
StarWitch: Sure, that wouldn't be too hard.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,179 @@
...
+            /opt/windows-x86-${{ env.taisei_version }}-setup-exe/Taisei-${{ env.taisei_version }}-setup-x86.exe.sig
+            /opt/emscripten-wasm32-${{ env.taisei_version }}-txz/Taisei-${{ env.taisei_version }}-emscripten-wasm32.tar.xz
+            /opt/emscripten-wasm32-${{ env.taisei_version }}-txz/Taisei-${{ env.taisei_version }}-emscripten-wasm32.tar.xz.sha256sum
+            /opt/emscripten-wasm32-${{ env.taisei_version }}-txz/Taisei-${{ env.taisei_version }}-emscripten-wasm32.tar.xz.sig
+            /opt/switch-aarch64-${{ env.taisei_version }}-zip/Taisei-${{ env.taisei_version }}-switch-aarch64.zip
+            /opt/switch-aarch64-${{ env.taisei_version }}-zip/Taisei-${{ env.taisei_version }}-switch-aarch64.zip.sha256sum
+            /opt/switch-aarch64-${{ env.taisei_version }}-zip/Taisei-${{ env.taisei_version }}-switch-aarch64.zip.sig
+            /opt/source-${{ env.taisei_version }}-txz/Taisei-${{ env.taisei_version }}-source.tar.xz
+            /opt/source-${{ env.taisei_version }}-txz/Taisei-${{ env.taisei_version }}-source.tar.xz.sha256sum
+            /opt/source-${{ env.taisei_version }}-txz/Taisei-${{ env.taisei_version }}-source.tar.xz.sig
```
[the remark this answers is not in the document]
StarWitch: We can probably make this a bit more procedural by switching the release action we use. I only went with this one as GitHub recommended it, but it seems like that recommendation is out of date and `softprops/action-gh-release` isn't maintained.

Here's a potential alternative: https://github.com/ncipollo/release-action

.github/workflows/publish.yml
```diff
@@ -0,0 +1,179 @@
...
+      - name: Set Env Vars
+        run: |
+          echo "taisei_version=$(echo '${{ github.event.inputs.tag }}' | sed 's/^v//g')" >> $GITHUB_ENV
+          # adds default changelog, which is blank, in case you forget so it doesn't hold up the whole release
+          # edit it on the actual GH release and commit what you write to master afterwards for posterity
+          if [ ! -f "changelogs/changelog-${{ github.event.inputs.tag }}.md" ]; then
+            echo "taisei_changelog_tag=default" >> $GITHUB_ENV
+          else
+            echo "taisei_changelog_tag=${{ github.event.inputs.tag }}" >> $GITHUB_ENV
+          fi
```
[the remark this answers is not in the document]
StarWitch: We can always pass in a blank changelog and then change the release "in post", so to speak. The idea with this is that if there _is_ a changelog we've written ahead of time, it can get used, but otherwise, it'll just fill it in with a blank one we can edit later. It's up to you, though.

.github/workflows/release.yml
StarWitch: Will do.

.github/workflows/publish.yml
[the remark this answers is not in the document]
StarWitch: Ended up going with the fork, for ease-of-use.

.github/workflows/publish.yml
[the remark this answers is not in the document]
StarWitch: Implemented within the build system now.

.github/workflows/publish.yml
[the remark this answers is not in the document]
StarWitch: Done. Much cleaner now.

StarWitch:
Spent the afternoon refactoring this. `sha256sum` and `sig` files are now generated by the new `releasefiles.py` script which `meson` now calls directly with `release_files` provided as an argument. I also found a better releasing script (the one I linked before) which allows wildcards, which is much simpler to use. I also forked the GPG action like you suggested, since that's safer/easier. Let me know if this is in the right direction.

StarWitch:
This PR is now a lot less cumbersome to review, as #321 was merged.

.github/workflows/release.yml
```diff
@@ -0,0 +1,556 @@
...
+    - name: Package (Binary)
+      run: ninja txz -C build/linux --verbose
+
+    - name: Upload Artifact (Binary)
+      uses: actions/upload-artifact@v2
+      if: success()
+      with:
+        name: linux-x86_64-${{ steps.vars.outputs.tag }}-txz
+        path: build/linux/Taisei-${{ steps.vars.outputs.tag }}-linux-x86_64.tar.xz*
+        if-no-files-found: error
```
Akaricchi, line 90: It looks like the `.sig` and `.sha256sum` files are never uploaded as artifacts. Am I missing something?

StarWitch: The `.sig` and `.sha256sum` are caught by the `*` at the end of the `path:`, as it gets generated as Taisei-...-x86_64.tar.gz.sha256sum`, etc.

.github/workflows/release.yml
```diff
@@ -0,0 +1,556 @@
...
+        apt install -y -qq software-properties-common
+
+        add-apt-repository ppa:git-core/ppa -y
+
+        apt install -y -qq
+        python3-docutils
+        python3-pip
+        git
+        wget
+        nsis
```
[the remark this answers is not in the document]
StarWitch: I think I originally did this so that every dependency that `nsis` depended on was installed before doing the raw `dpkg -i`, but I'll double-check that.

scripts/taiseilib/version.py
```diff
@@ -68,7 +68,9 @@ def get(*, rootdir=None, fallback=None, args=common.default_args):
             universal_newlines=True
         ).strip()
 
-        if '-' in version_str:
+        # does not append -HEAD if the release is tagged -beta, -rc, etc.
```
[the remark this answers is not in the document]
StarWitch: This is due to an unfortunate long-standing bug in GitHub CI: https://github.com/actions/checkout/issues/290

.github/workflows/release.yml
```diff
@@ -0,0 +1,556 @@
...
+      uses: actions/upload-artifact@v2
+      if: success()
+      with:
+        name: linux-x86_64-${{ steps.vars.outputs.tag }}-txz
+        path: build/linux/Taisei-${{ steps.vars.outputs.tag }}-linux-x86_64.tar.xz*
+        if-no-files-found: error
+
+    - name: Package (Source)
+      run: >
+        meson rewrite kwargs set project / version v${{ steps.vars.outputs.tag }}
```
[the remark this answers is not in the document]
StarWitch: `meson dist` goes off of the `project.version` set in `meson.build`, which then outputs files labeled `v1.4-dev` (currently) instead of whatever tag is being used in `git`. There might be some way around it, but this `rewrite` I have in there is intended as a workaround which then uses the tag from git, as expected.

.github/workflows/release.yml
```diff
@@ -0,0 +1,556 @@
...
+
+        meson setup build/emscripten
+        --cross-file misc/ci/emscripten-build.ini
+        --cross-file misc/ci/emscripten-ephemeral-ci.ini
+        --prefix=$(pwd)/build-release
+        -Drelease_files=true
+
+        meson configure
+        -Dbuild.cpp_std=gnu++14
+        build/emscripten # set correct C++ version
```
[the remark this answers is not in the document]
StarWitch: Seems to work just fine. Removed.

StarWitch:
Changes are in, and builds are now passing: https://github.com/StarWitch/taisei/runs/5260920968

scripts/taiseilib/integrityfiles.py
```diff
@@ -0,0 +1,66 @@
...
+
+        sig.close()
+    with open(archive, 'rb') as file:
+        with open('{0}.sha256sum'.format(archive), 'r') as sha256file:
+            sha256input = sha256file.readline().strip()
+            sha256output = '{0}  {1}'.format(hashlib.sha256(file.read()).hexdigest(), archive)
+            if str(sha256input) != str(sha256output):
+                raise ValueError('could not verify sha256sum on archive: {0}, mismatch: {1} != {2}'.format(archive, sha256input, sha256output))
+            sha256file.close()
+        file.close()
```
Akaricchi, line 50: Could benefit a lot from pathlib refactoring

```python
archive = Path(archive)
sha256file = Path('{0}.sha256sum'.format(archive))
sha256input = sha256file.read_text().strip()
sha256output = '{0}  {1}'.format(hashlib.sha256(archive.read_bytes()).hexdigest(), archive.name)
```

meson_options.txt
```diff
@@ -197,3 +197,10 @@ option(
     value : 'false',
     description : 'Enable live-reloading workflow for stages (for development only)'
 )
+
+option(
+    'integrity_files',
+    type : 'boolean',
+    value : false,
+    description : 'Generate .sig and .sha256sum files of built archives for releases'
+)
```
Akaricchi, line 206: To be honest I don't quite like polluting the list of options with this, if it can be avoided. Maybe we should just call `scripts/sign-release.py` from CI? That would simplify things a bit on the build system side.

scripts/taiseilib/version.py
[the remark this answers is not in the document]
Akaricchi: That seems pretty bad. Have you tried any of the workarounds suggested in that issue, such as `git fetch --force --tags`?

meson_options.txt
StarWitch: All done, it's manually run through the script now.

StarWitch:
New changes in and builds working: https://github.com/StarWitch/taisei/actions/runs/1872514485

StarWitch:
Updated to use the new custom Ubuntu 20.04/llvm-mingw image.

Here's it in action: https://github.com/StarWitch/taisei/actions/runs/1873250975

Cuts off about 2-3 minutes from the build, and means a few stupid workarounds I had for Python3.6 could be removed.

.github/workflows/release.yml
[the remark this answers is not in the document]
eli-schwartz: But then you'll still have the tarball's copy of `meson.build` using the outdated version number?

Really, the version number should be incremented in an actual commit, and that commit is what should be tagged.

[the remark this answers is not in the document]
Akaricchi: > But then you'll still have the tarball's copy of `meson.build` using the outdated version number?
> 
> Really, the version number should be incremented in an actual commit, and that commit is what should be tagged.

For an actual release I 100% agree with you. But this wouldn't work for a quick test run, which I presume this workaround is for. It is really dirty though; I'm quite unhappy with it. I didn't outright object to it because I don't have a better idea right now.

.github/workflows/release.yml
```diff
@@ -0,0 +1,488 @@
...
+        python3-docutils
+        gnupg
+
+        pip install
+        meson==${{ env.MESON_VERSION }}
+        ninja
+        zstandard
+        python-gnupg
+
+    # NOTE: confusingly, 'fetch-depth: 0' == "get all history", not "no history"
```
Akaricchi, line 45: It's the same as the `--depth` argument to `git clone`. 1 is "one commit of history" (the latest); "no history" wouldn't make sense.

.github/workflows/release.yml
```diff
@@ -0,0 +1,488 @@
+name: "Release Builds"
+on:
+  push:
+    branches:
+      - v[0-9].[0-9]**
+    tags:
+      - v[0-9].[0-9]**
+      - v[0-9].[0-9]**-**
```
Akaricchi, line 8: Would it be possible to manually run this on some arbitrary commit?

Akaricchi:
I've added [a new test script](https://github.com/taisei-project/taisei/commit/9669e1d30a665c3f1a2ffc856d89756aaef916d1) today. It should be ran after the linux standalone binary is built, and the workflow should fail if it returns non-zero. Can you please rebase this to latest master and incorporate that change?

.github/workflows/release.yml
StarWitch: Do you mean from a security standpoint, or a convenience standpoint?

In terms of convenience, I can set up a `workflow_dispatch` condition and have it provide a field you can enter to select a specific commit. It would then check out that commit and build all the binaries as usual.

In terms of security, there's a blurb at the top of the `Secrets` page:

>Secrets are environment variables that are encrypted. Anyone with collaborator access to this repository can use these secrets for Actions.

>Secrets are not passed to workflows that are triggered by a pull request from a fork. [Learn more](https://docs.github.com/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets).

So as long as the list of _collaborators_ is kept managed appropriately, using the `Secrets` in workflows is relatively safe; people can't just fork Taisei and run an arbitrary workflow to get the secrets.

.github/workflows/release.yml
StarWitch: Right, but that wasn't intuitive to me from the name of the option, hence the comment.

.github/workflows/release.yml
Akaricchi: From a convenience standpoint; I've read up on security already.

> In terms of convenience, I can set up a workflow_dispatch condition and have it provide a field you can enter to select a specific commit. It would then check out that commit and build all the binaries as usual.

Sounds good to me. Maybe later we can also upgrade this to provide e.g. nightly autobuilds.

StarWitch:
A bunch of changes since I last updated this PR.

1. Updated all of the action versions to get rid of as many deprecation warnings as possible.
2. Updated to `ubuntu-20.04` since `ubuntu-18.04` is being removed from GitHub Actions in six months. We could roll our own but it was deemed not worth it.
3. Updated `macos-11` to `macos-12`.
4. Allow specifying of an arbitrary commit through `workflow_dispatch`. 
5. Added the Linux library linker checking script to the Linux stage. 
6. Some minor optimizations.

.github/workflows/release.yml
```diff
@@ -0,0 +1,576 @@
...
+        --prefix=$(pwd)/build-release
+
+    - name: Set Package Version
+      run: |
+        if git describe --exact-match &>/dev/null && ${{ (github.event.inputs.custom-ref && false) || true }}
+        then
+            echo "BUILD_VERSION=${GITHUB_REF#refs/tags/v}" >> $GITHUB_ENV
+        else
+            echo "BUILD_VERSION=$(meson introspect --projectinfo build/linux | jq -r '.version' | cut -c 2-)" >> $GITHUB_ENV
+        fi
```
Akaricchi, line 88: (not a complete review yet)

It seems like the purpose of this variable is to "guess" the names of the release artifacts? I don't think this will work with custom refs, at least not in all cases. Regardless, we have a much better way of doing this already: just call `scripts/version.py`. That's what the build system uses to construct those file names.

StarWitch: Done. Turns out `scripts/version.py` does do what I need. 

Currently macOS release builds are failing due to [this bug](https://github.com/actions/runner-images/issues/6507) but that should be fixed in the next day or two. I'll re-run [this build](https://github.com/StarWitch/taisei/actions/runs/3406325345/jobs/5664932541) to test it when it's merged.

Akaricchi (commented):
Everything looks good to land modulo comments. I hope this is my last review here. We can keep improving it for a long time, but this PR has been in the oven for way too long already. Great work!

StarWitch:
Remaining comments resolved.

There's currently an issue with macOS builds due to [this issue](https://github.com/actions/runner-images/issues/6507), but the fix for that was merged and it should be deployed to the runner pool soon, which shouldn't require editing the workflow at all.

Akaricchi (approved):

PR #291, opened 2021-05-03
enemy: extended events

Akaricchi:
This is based on similar work done for projectiles in f25c8894. Enables interception of damage events and ability to determine the cause of death.

This also subtly changes the scheduling of 'killed' events; they are fired as soon as the death is registered now instead of just before deletion.


Akaricchi:
Merged in 7da2c7e5adced2c2882abff78466c3031aae85f6

PR #292, opened 2021-05-03
Makefile

StarWitch:
Basic Makefile, used to document common commands.

Depends on #278 and #289.


StarWitch:
Closing this as I don't think it's really needed with the documentation I've written, and it's better not to hide the process from people.

PR #295, opened 2021-05-06
Add -lwebgl.js to fix macOS Emscripten build error

StarWitch:
Simple one-line fix.

See: https://github.com/emscripten-core/emscripten/issues/10171


Akaricchi:
The amount of flags required to get webgl working is getting ridiculous...

PR #296, opened 2021-05-07
Emscripten CI test builds

StarWitch:
Provides a simple Emscripten test build to make sure the damn thing still compiles with its arcane toolchain.

Also adds:
* prettier labels for the tests themselves (i.e: `Linux Test Build (CI)`)
* moves to meson `0.56.2`
* fixes a couple of syntax errors in the machine files
* turns `werror` back on for all builds

See: #289

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,9 +20,8 @@ on:
       - "**.yaml"
       - "**.build"
 
-# TODO: use later fixed version of emscripten when it releases
 env:
-  EM_VERSION: '2.0.19'
+  EM_VERSION: '2.0.21'
   EM_CACHE_FOLDER: 'emsdk'
   TAISEI_NOPRELOAD: 0
   TAISEI_PRELOAD_REQUIRED: 1
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,11 +1,11 @@
-name: Taisei Test Build (CI)
+name: Test Builds
 on:
   push:
     branches:
       - master
-      - v1.*
+      - v*.*
     tags:
-      - v1.*
+      - v*.*
     paths:
       - "**.c"
       - "**.h"
@@ -24,12 +24,13 @@ env:
   EM_VERSION: '2.0.21'
   EM_VERSION_EXPERIMENTAL: 'latest'
   EM_CACHE_FOLDER: 'emsdk'
+  MESON_VERSION: '0.56.2'
   TAISEI_NOPRELOAD: 0
   TAISEI_PRELOAD_REQUIRED: 1
 
 jobs:
   linux-test-build:
-    name: Linux Test (x64)
+    name: Linux (x64)
     if: "!contains(github.event.head_commit.message, '[skip ci]')"
     runs-on: ubuntu-20.04
     steps:
@@ -45,7 +46,7 @@ jobs:
 
     - name: Install Tools
       run: >
-        sudo apt update || true
+        sudo apt update || true # debian cache is not atomic and can sometimes fail
 
         sudo apt install -y -qq
         build-essential
@@ -62,7 +63,7 @@ jobs:
         libzstd-dev
 
         pip install
-        meson==0.56.2
+        meson==${{ env.MESON_VERSION }}
         ninja
         zstandard
 
@@ -77,7 +78,7 @@ jobs:
         meson compile -C build/ --verbose
         meson install -C build/
 
-    - name: Test (Taisei)
+    - name: (Taisei)
       run: $(pwd)/build-test/bin/taisei -R $(pwd)/misc/ci/tests/test-replay.tsr
       env:
         TAISEI_NOPRELOAD: ${{ env.TAISEI_NOPRELOAD }}
@@ -91,7 +92,7 @@ jobs:
         if-no-files-found: warn
 
   macos-test-build:
-    name: macOS Test (x64)
+    name: macOS (x64)
     if: "!contains(github.event.head_commit.message, '[skip ci]')"
     runs-on: macos-10.15
     steps:
@@ -121,7 +122,7 @@ jobs:
         ninja
 
         pip install
-        meson==0.56.2
+        meson==${{ env.MESON_VERSION }}
         ninja
         zstandard
 
@@ -136,7 +137,7 @@ jobs:
         meson compile -C build/ --verbose
         meson install -C build/
 
-    - name: Test (Taisei)
+    - name: (Taisei)
       run: $(pwd)/build-test/Taisei.app/Contents/MacOS/Taisei -R $(pwd)/misc/ci/tests/test-replay.tsr
       env:
         TAISEI_NOPRELOAD: ${{ env.TAISEI_NOPRELOAD }}
@@ -150,7 +151,7 @@ jobs:
         if-no-files-found: warn
 
   windows-test-build:
-    name: Windows Test (x64)
+    name: Windows (x64)
     if: "!contains(github.event.head_commit.message, '[skip ci]')"
     runs-on: ubuntu-20.04
     container: mstorsjo/llvm-mingw:20210423 # cross-compiler with mingw
@@ -170,7 +171,7 @@ jobs:
         nsis
 
         pip3 install
-        meson==0.56.2
+        meson==${{ env.MESON_VERSION }}
         ninja
         zstandard
 
@@ -198,7 +199,7 @@ jobs:
         if-no-files-found: warn
 
   emscripten-test-build:
-    name: "Emscripten Test (emsdk: stable) (WebGL)"
+    name: "Emscripten (emsdk: stable)"
     if: "!contains(github.event.head_commit.message, '[skip ci]')"
     runs-on: ubuntu-20.04
     steps:
@@ -222,7 +223,7 @@ jobs:
         git
 
         pip install
-        meson==0.56.2
+        meson==${{ env.MESON_VERSION }}
         ninja
         zstandard
 
@@ -247,7 +248,7 @@ jobs:
         emcc -v
         tee misc/ci/emscripten-ephemeral-ci.ini <<EOF >/dev/null
         [constants]
-        toolchain = '$(pwd)/emsdk/upstream/emscripten/'
+        toolchain = '$(pwd)/${{ env.EM_CACHE_FOLDER }}/upstream/emscripten/'
         EOF
 
     - name: Configure
@@ -278,7 +279,7 @@ jobs:
         if-no-files-found: warn
 
   emscripten-test-build-latest:
-    name: "Emscripten Test (emsdk: experimental) (WebGL)"
+    name: "Emscripten (emsdk: latest)"
     if: "!contains(github.event.head_commit.message, '[skip ci]')"
     runs-on: ubuntu-20.04
     continue-on-error: true
@@ -303,7 +304,7 @@ jobs:
         git
 
         pip install
-        meson==0.56.2
+        meson==${{ env.MESON_VERSION }}
         ninja
         zstandard
 
@@ -319,7 +320,7 @@ jobs:
         if [[ ! -d ./emsdk ]]; then
           git clone https://github.com/emscripten-core/emsdk.git
         else
-          bash -c "cd emsdk; git pull"
+          bash -c "cd emsdk; git fetch --all; git reset --hard origin/master"
         fi
         emsdk/emsdk install ${{ env.EM_VERSION_EXPERIMENTAL }} || true
         emsdk/emsdk activate ${{ env.EM_VERSION_EXPERIMENTAL }}
@@ -330,7 +331,7 @@ jobs:
         emcc -v
         tee misc/ci/emscripten-ephemeral-ci.ini <<EOF >/dev/null
         [constants]
-        toolchain = '$(pwd)/emsdk/upstream/emscripten/'
+        toolchain = '$(pwd)/${{ env.EM_CACHE_FOLDER }}/upstream/emscripten/'
         EOF
 
     - name: Configure
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,10 +21,9 @@ on:
       - "**.build"
 
 env:
-  EM_VERSION: '2.0.21'
-  EM_VERSION_EXPERIMENTAL: 'latest'
-  EM_CACHE_FOLDER: 'emsdk'
   MESON_VERSION: '0.56.2'
+  EM_VERSION: '2.0.21'
+  EM_CACHE_FOLDER: 'emsdk'
   TAISEI_NOPRELOAD: 0
   TAISEI_PRELOAD_REQUIRED: 1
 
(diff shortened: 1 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -186,3 +196,78 @@ jobs:
...
+        meson configure
+        -Dbuild.cpp_std=gnu++14
+        build/
+
+    - name: Build
+      run: |
+        source emsdk/emsdk_env.sh
+        # run twice due to a header race condition with SPIRV-Tools
+        meson compile txz -C build/ --verbose || true
+        meson compile txz -C build/ --verbose
```
Akaricchi, line 266: Can you elaborate on what's going on here? Sounds like something we definitely want fixed.

Akaricchi: Side note, might be better to use `tar` instead of `txz` here. The `xz` compression is quote slow and doesn't reduce the file size much here (and we ignore the artifact, anyway).

.github/workflows/main.yml
```diff
@@ -186,3 +196,78 @@ jobs:
...
+        meson==0.56.2
+        ninja
+        zstandard
+
+    - name: Pull cached emsdk
+      id: cache-emsdk
+      uses: actions/cache@v2
+      with:
+        path: ${{ env.EM_CACHE_FOLDER }}
+        key: ${{ env.EM_VERSION }}-${{ runner.os }}
```
Akaricchi, line 234: The cache mechanism is interesting. Maybe we could build and install the native shader tools the same way (spirv-cross and glslc)? Those probably take more time to build than Taisei itself. Doesn't have to be done in this PR, though.

misc/ci/emscripten-build-test-ci.ini
```diff
@@ -0,0 +1,53 @@
+[constants]
+cflags = []
+ldflags = ['-v']
+
+[binaries]
+ar = 'emar'
+c = 'emcc'
+cpp = 'em++'
+ranlib = 'emranlib'
+file_packager = '/home/runner/work/taisei/taisei/emsdk/upstream/emscripten/tools/file_packager'
```
Akaricchi, line 10: I'd rather put the toolchain path in constants, and be explicit with all the binaries:

```ini
[constants]
toolchain = '/home/runner/work/taisei/taisei/emsdk/upstream/emscripten'
cflags = []
ldflags = ['-v']

[binaries]
ar = toolchain / 'emar'
c = toolchain / 'emcc'
cpp = toolchain / 'em++'
ranlib = toolchain / 'ranlib'
file_packager = toolchain / 'tools/file_packager'
```

(Note that unquoted `/` is Meson's path-joining operator here)

This setup doesn't rely on emscripten being in `PATH`.

As a further improvement, we can omit the toolchain path from the static crossfile and leverage Meson's layering feature to inject it at CI runtime. This makes it possible to use the crossfile unmodified in other environments.

Pseudo-code:

```sh
cat <<EOF
[constants]
toolchain = '${MY_TOOLCHAIN_PATH}'
EOF > emscripten-path-ephemeral.ini
meson setup \
    --cross-file emscripten-path-ephemeral.ini \
    --cross-file misc/ci/emscripten-build-test-ci.ini \
    source-dir build-dir
```

.github/workflows/main.yml
StarWitch: It's a bit trickier to pull off with things inside the `build/` directory simply because of how picky `meson` is about build caches. It may or may not save time in the end.

But yeah, let's do that in another PR.

.github/workflows/main.yml
StarWitch: For the error, here's the log: https://github.com/StarWitch/taisei/runs/2522851326?check_suite_focus=true

```
 FAILED: subprojects/SPIRV-Tools/source/opt/libSPIRV-Tools-opt-native.a.p/pch.h.pch 
c++ -Isubprojects/SPIRV-Tools/source/opt/libSPIRV-Tools-opt-native.a.p -Isubprojects/SPIRV-Tools/source/opt -I../subprojects/SPIRV-Tools/source/opt -Isubprojects/SPIRV-Tools -I../subprojects/SPIRV-Tools -Isubprojects/SPIRV-Tools/source -I../subprojects/SPIRV-Tools/source -I../subprojects/SPIRV-Tools/include -I../subprojects/SPIRV-Headers/include -Xclang -fcolor-diagnostics -DNDEBUG -pipe -Wall -Winvalid-pch -Wnon-virtual-dtor -std=gnu++14 -x c++-header -MD -MQ subprojects/SPIRV-Tools/source/opt/libSPIRV-Tools-opt-native.a.p/pch.h.pch -MF subprojects/SPIRV-Tools/source/opt/libSPIRV-Tools-opt-native.a.p/pch.h.pch.d -o subprojects/SPIRV-Tools/source/opt/libSPIRV-Tools-opt-native.a.p/pch.h.pch -c ../subprojects/SPIRV-Tools/source/opt/meson-pch/pch.h
In file included from ../subprojects/SPIRV-Tools/source/opt/meson-pch/pch.h:24:
In file included from ../subprojects/SPIRV-Tools/source/opt/basic_block.h:29:
../subprojects/SPIRV-Tools/source/opt/instruction.h:25:10: fatal error: 'OpenCLDebugInfo100.h' file not found
#include "OpenCLDebugInfo100.h"
         ^~~~~~~~~~~~~~~~~~~~~~
```

Basically, the build gets ahead of itself, and can occasionally fail to find this file because it isn't in place yet. Running the command again fixes it.

.github/workflows/main.yml
Akaricchi: It's not necessary (or recommended) to mess with the build directory here. We can just install the correct versions of the tools into an isolated (cached) directory, disable the `force_vendored_shader_tools` option, and point Meson to the prebuilt binaries (either via `PATH` manipulation or explicitly in machine files)

.github/workflows/main.yml
Akaricchi: Seems like SPIRV-Tools's internal dependencies aren't set up adequately. I'll look into this soon.

.github/workflows/main.yml
```diff
@@ -37,7 +45,7 @@ jobs:
 
     - name: Install Build Tools (apt)
       run: >
-        sudo apt update
+        sudo apt update || true
```
Akaricchi, line 48: What's up with that?

.github/workflows/main.yml
StarWitch: Cool. If you don't have time to investigate/fix it right now, this workaround does work for now. We can always remove it when you commit that change.

.github/workflows/main.yml
StarWitch: Sometimes, `apt update` will catch a mirror in the middle of a mirror sync and spit out a useless error. It's mostly harmless, since the apt cache is still valid and you can still install packages (especially since GitHub keeps their images up to date), but it does throw an `exit 1` and error out the CI job. It's a rare thing to happen but it has happened once or twice while testing CI, so I figured it was worth a quick workaround to avoid it.

Akaricchi: Ah, so it's debian being garbage as usual, ok.

> since the apt cache is still valid and you can still install packages (especially since GitHub keeps their images up to date)

Does this mean the images already come with a reasonably up to date apt cache? If so, do we even need `apt update` at all?

StarWitch: It's still a good practice to do it, since sometimes a random package will be deleted from their download stores, i.e: in the case of a severe CVE where they don't want anyone downloading it if possible. It's doubly-rare that both `apt update` *and* a package would be deleted, so I think it's fine for now.

.github/workflows/main.yml
```diff
@@ -1,4 +1,4 @@
-name: taisei-build-tests
+name: Taisei Test Build (CI)
```
Akaricchi, line 1: Minor nitpick, but I'm not a fan of the verbose names. It looks overloaded, and one of the entries even obscures the build time:
image

* It is obvious that we're building Taisei, so that part can be dropped
* Likewise, the "(CI)" offers no useful information
* Since we have "Test Build" in the suite name, we don't need to restate "Test" in each run
* Emscripten builds always use WebGL, so we don't need to state that either
* "emsdk" can be dropped, just the emscripten version in parentheses is enough
    * On that note, I think "experimental" should be replaced with "latest" since that's the tag they're using
    * Not sure if possible, but it'd be nice to have the actual version (i.e. the value of `EM_VERSION` from `env`) instead of "stable"

.github/workflows/main.yml
Akaricchi: A brief comment explaining the purpose of it would be nice, then.

.github/workflows/main.yml
```diff
@@ -54,26 +62,26 @@ jobs:
         libzstd-dev
 
         pip install
-        meson==0.55.3
+        meson==0.56.2
```
Akaricchi, line 65: Maybe the pinned meson version should also go to `env`?

.github/workflows/main.yml
```diff
@@ -186,3 +196,167 @@ jobs:
...
+        ninja
+        zstandard
+
+    - name: Pull cached emsdk
+      id: cache-emsdk
+      uses: actions/cache@v2
+      with:
+        path: ${{ env.EM_CACHE_FOLDER }}
+        key: ${{ env.EM_VERSION_EXPERIMENTAL }}-${{ runner.os }}
```
Akaricchi, line 315: How is this going to work after emscripten updates? Won't we have a stale cache entry for `latest-whateverOS` with the previous version? Is the cache invalidated after some time?

StarWitch: It's still useful to pull the cache, but the command below makes sure to fetch and install the latest version anyway, in case it's been updated. It speeds things up when a new version hasn't dropped.

.github/workflows/main.yml
StarWitch: Done!

.github/workflows/main.yml
StarWitch: Done.

.github/workflows/main.yml
Akaricchi: Ok, so it'll install the latest version. But what happens on subsequent builds? Does it pull the old version from the cache and install the new one over it again? Or does it somehow eventually update/invalidate the cache entry?

.github/workflows/main.yml
```diff
@@ -77,7 +78,7 @@ jobs:
         meson compile -C build/ --verbose
         meson install -C build/
 
-    - name: Test (Taisei)
+    - name: (Taisei)
```
Akaricchi: That's a weird name for a step. Overzealous substitution? Maybe call it "run tests" or something.

.github/workflows/main.yml
```diff
@@ -319,7 +320,7 @@ jobs:
         if [[ ! -d ./emsdk ]]; then
           git clone https://github.com/emscripten-core/emsdk.git
         else
-          bash -c "cd emsdk; git pull"
+          bash -c "cd emsdk; git fetch --all; git reset --hard origin/master"
```
Akaricchi, line 323: Did they not switch this one to `main`? They did it for the emscripten repo, but i don't know about emsdk. 

(Did I mention how much i fucking hate this branch-renaming braindeath?)

.github/workflows/main.yml
StarWitch: Done. I'd actually moved to this short form in the release pipeline branch, but forgot to port those changes over to these tests. 

As for the actual version being in the name, I tried a bunch of ways of getting the version number into the actual name of the test, but unfortunately it's not possible throw any `env` vars into it for whatever reason. 

One way of doing it would be to use a `strategy: matrix:` which would allow us to access `{{ matrix.* }}` variables (allowing us to define `matrix.emsdk`, for example), but while that *is* a lot more powerful, it's also a bit overkill for what we've got going on here, and would be a whole big change all on its own. Maybe if we decide we really want to test against 6 different versions of Ubuntu (which is pretty much what `matrix` is intended for), then it'd be worth it.

.github/workflows/main.yml
StarWitch: They did, and it's even more awful because `origin/master` still resolves correctly, even though it's not valid. 😞

Akaricchi: It's "valid" in that it refers to an existing branch in the remote… they didn't bother deleting the old branch, and of course it's not being updated. Way to guarantee a smooth transition.

StarWitch: Yep. It sucks.

.github/workflows/main.yml
StarWitch: The `latest` in `./emsdk install latest` is actually a tag they change in `main`, as seen here: https://github.com/emscripten-core/emsdk/blob/32fd1fafac571965ce88f778dac1072ed0d7e061/emscripten-releases-tags.txt#L2

If `latest` doesn't change when the step below this one `reset`s against `origin/main`, `install` and `activate` basically do nothing. If `latest` *has* changed, then it'll install (and cache) that new version. It still speeds up the process if the `latest` tag hasn't changed.

You can test it out by doing something like:
```
git checkout 2.0.20
./emsdk install latest
git checkout 2.0.21
./emsdk install latest
```

You'll see that it actually pulls new versions each time. This simulates what would happen if `main` were to change as well.

Akaricchi: > If `latest` has changed, then it'll install **(and cache)** that new version.

The only thing that's not clear to me is the part in bold. Are we *sure* it will actually cache the new version in this case? Does it *always* update the cached entry, even after a cache hit? That's the only way this setup can work properly, so we need to make sure of that. Though in this case it acts less like a cache and more like a mutable persistent filesystem. Which is a strange thing to have in a controlled CI environment, so it's reasonable to expect that it does not work like that.

StarWitch: Unfortunately, you're correct. Good catch.

I had thought that if the cache directory had changed, it would update it with the final cleanup action, but apparently cache hits are immutable. I seriously wonder what GitHub's design philosophy was with their CI, it feels so undercooked.

Akaricchi: I think it's not necessarily a bad design choice. Having a persistent read-write filesystem equivalent in a supposedly reproducible build setup doesn't make a lot of sense (but then neither does having direct internet access, so i guess it's a moot point).

Anyway, the real design flaw is that it's not possible to compute the cache key dynamically in a previous step. Or is it? I don't actually know. But if it is possible, that is what we actually want: figure out what the latest version is, make a cache key out of it, try to get it from the cache, if missed then install it.

If that's not possible, then I don't think having a test build for the latest version is a good idea. I'm not comfortable with reinstalling emscripten from scratch on every push. It's reasonable to just say we only support one emscripten version — the one we test against — since they break it so often, and just bump it manually occasionally. This is kind of what I've been doing already anyway, except no one knew what emscripten version we're actually targeting.

StarWitch: Alright, what I've done in the latest commit is explicitly pull the latest tag out of `emsdk` using `git describe`. It then uses that tag to fetch that version from the cache, and if it isn't found, it `install`/`activate`s. Also, if `EM_VERSION` == `latest`, then the other steps in the job get skipped entirely.

Akaricchi (approved):

PR #303, opened 2021-06-18
Improved README.

StarWitch:
* Expands upon what both Taisei and Touhou are
* Removes technical stuff that will be added in future PRs (game controller, etc)
* Removes freenode link (RIP)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -100,10 +100,11 @@ Optional
 Build-Time Dependenices
 _______________________
 
+-  ``gcc`` or ``clang``
+-  meson >= 0.53.0
 -  Python >= 3.6
 -  `python-zstandard <https://github.com/indygreg/python-zstandard>`__ >= 0.11.1
--  meson >= 0.53.0
--  ``gcc`` or ``clang``
+-  `python-docutils <https://pypi.org/project/docutils/>`__ (optional, for generating documentation)
 
 Obtaining Source Code
 ^^^^^^^^^^^^^^^^^^^^^
--- a/README.rst
+++ b/README.rst
@@ -6,18 +6,102 @@ Taisei
 Introduction
 ------------
 
-Taisei is an open clone of the Tōhō Project series. Tōhō is a one-man project of
-shoot-em-up games set in an isolated world full of Japanese folklore.
+About Taisei Project
+^^^^^^^^^^^^^^^^^^^^
+
+Taisei Project is an open source fan-game set in the world of Tōhō
+Project. It is a top-down vertical-scrolling curtain fire shooting game (STG),
+also known as a "bullet hell" or "danmaku," with an original soundtrack,
+character art, and story. It is a fast-paced game focused around pattern
+recognition and mastery through practice.
+
+Taisei Project is highly portable, and is written in C (specifically, C11),
+using SDL2 with an OpenGL renderer. It is officially supported on Windows,
+Linux, macOS, and through WebGL-enabled browsers such as Firefox and
+Chromium-based browsers (Chrome, Edge, etc). It can also be compiled for a
+number of other operating systems.
+
+For gameplay instructions, read `this <doc/GAME.rst>`__.
+
+For the story, read `this <doc/STORY.txt>`__. (Spoiler warning!)
+
+About Tōhō Project
+^^^^^^^^^^^^^^^^^^
+
+Tōhō Project, the series Taisei Project is based upon, is a modern fairytale
+series set in Gensōkyō, or the "Land of Fantasy," a pocket dimension hidden away
+in modern-day Japan. It primarily focuses on Japanese folklore and myths,
+starring a ragtag group of women who keep the peace among all of the
+supernatural inhabitants.
+
+Tōhō Project is known for its ensemble cast of characters and memorable
+soundtracks. It is an indie series (also called "doujin" in Japanese) produced,
+by and large by a single artist known as ZUN, and has a
+`permissive license <https://en.touhouwiki.net/wiki/Touhou_Wiki:Copyrights#Copyright_status.2FTerms_of_Use_of_the_Touhou_Project>`__
+which allows for indie derivative works such as Taisei Project to legally exist.
+
+For more information on Tōhō Project,
+`click here <https://en.wikipedia.org/wiki/Touhou_Project>`__.
+
+For more information on dōjin culture,
+`click here <https://en.wikipedia.org/wiki/D%C5%8Djin>`__.
 
 Installation
 ------------
 
-Dependencies
-^^^^^^^^^^^^
+You can find precompiled binaries of the complete game on the
+`Releases <https://github.com/taisei-project/taisei/releases>`__ page on
+GitHub, available for Windows (x86/x64), Linux, and macOS.
 
--  OpenGL >= 3.3 or OpenGL ES >= 3.0 or OpenGL ES >= 2.0 (with some extensions)
--  `cglm <https://github.com/recp/cglm>`__ >= 0.7.8
+An experimental build for Nintendo Switch (homebrew) also exists (use at your
+own risk).
+
+You can also play our experimental WebGL build through your web browser
+`here <https://play.taisei-project.org/>`__. (Chromium-based browsers and
+Firefox supported.)
+
+Source Code & Development
+-------------------------
+
+**Important:** As of v1.3, you cannot download GitHub's automatically-generated
+source archives, as they don't contain the necessary submodules for compiling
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -6,126 +6,122 @@ Taisei
...
-- ``OES_vertex_array_object``
-- ``EXT_frag_depth``
-- ``EXT_instanced_arrays`` or ``ANGLE_instanced_arrays`` or
-  ``NV_instanced_arrays``
+-  OpenSSL (for a better SHA-256 implementation; used in shader cache)
+-  SPIRV-Cross >= 2019-03-22 (for OpenGL ES backends)
+-  libshaderc (for OpenGL ES backends)
+-  ANGLE (for ANGLE support, required for Windows)
+-  GameMode headers (Linux only; for automatic `GameMode
+   <https://github.com/FeralInteractive/gamemode>`__ integration)
```
Akaricchi, line 120: Could of things here

* I don't like the merging of runtime dependencies with build-time dependencies. The distinction is pretty important particularly in a cross-compilation environment.
* cglm is now missing (runtime dependency)
* python-zstandard is also missing (build dependency)
* ANGLE is not "required" on windows. For the purposes of hacking on the codebase, you probably won't even need it, unless you're stuck with an Intel iGPU.
* I think we can just remove mentions of OpenGL ES 2.0, and simply state that GL3.3+ or ES3.0+ is required. The gles20 backend is unmaintained and doesn't really work; at best it'll only work on ES3-capable hardware.

README.rst
```diff
@@ -6,126 +6,122 @@ Taisei
...
 
-The OpenGL ES 3.0 backend is not built by default. To enable it, do:
+If you want to pull them in manually, be sure to run the following:
 
 ::
 
-    meson configure -Dr_gles30=true -Dshader_transpiler=true
+    cd taisei
+    git submodule init
+    git submodule update
```
Akaricchi, line 89: Actually you can just do `git submodule update --init` with non-ancient git.

README.rst
```diff
@@ -6,126 +6,122 @@ Taisei
...
+
+You should also run ``git submodule update`` whenever you pull in
+new code, checkout another branch, etc. The ``pull`` and ``checkout`` helper
+scripts can do that for you automatically.
 
-See `here <doc/ENVIRON.rst>`__ for information on how to activate it.
-Alternatively, do this to make GLES 3.0 the default backend:
+For more in-depth development and compiling instructions, see our
+`Development Guide <doc/DEVELOPMENT.rst>`__.
```
Akaricchi, line 96: I think it's very important to have a super basic TL;DR version of how to build and run the thing, i.e. literally
```sh
meson setup build
meson compile -C build
meson install -C build
```
Asking a potential new user/contributor to read a book that doesn't even exist yet is not the best look. On that note, I'd also remove references to not-yet-written documents, and have them added back in when they are ready.

Would be even better if all the info on how to properly obtain the source code could be further condensed somehow. Perhaps it's not necessary to mention the alternative way to init submodules… teaching people how git works is not a goal here.

README.rst
```diff
@@ -6,126 +6,122 @@ Taisei
...
 
-Build-only dependencies
-^^^^^^^^^^^^^^^^^^^^^^^
-
--  Python >= 3.6
--  meson >= 0.53.0
--  [python-zstandard](https://github.com/indygreg/python-zstandard)
-
-Optional:
+Taisei Project is highly portable, and is written in C (specifically, C11),
```
Akaricchi, line 18: Probably can just say written in C11.

StarWitch:
Fixed all the PR comments, and added the icon to the header too. Let me know how it looks.

The further dev instructions/troubleshooting links will get added back in future PRs.

README.rst
```diff
@@ -1,131 +1,156 @@
...
 -  zlib
 
-Optional:
+Optional
+********
 
 -  OpenSSL (for a better SHA-256 implementation; used in shader cache)
 -  SPIRV-Cross >= 2019-03-22 (for OpenGL ES backends)
 -  libshaderc (for OpenGL ES backends)
+-  `ANGLE <https://github.com/google/angle>`__ (for ANGLE support, useful for Windows builds)
```
Akaricchi, line 94: `ANGLE (for ANGLE support)` is pretty tautological. Maybe better to say that it's usable via the gles30 backend and/or that it's for rendering on non-OpenGL platforms?

README.rst
```diff
@@ -1,131 +1,156 @@
...
-Build-only dependencies
-^^^^^^^^^^^^^^^^^^^^^^^
+Build-Time Dependenices
+_______________________
 
 -  Python >= 3.6
--  meson >= 0.53.0
--  [python-zstandard](https://github.com/indygreg/python-zstandard)
+-  `python-zstandard <https://github.com/indygreg/python-zstandard>`__ >= 0.11.1
+-  meson >= 0.56.0
```
Akaricchi, line 103: We still support >=0.53.0 for regular builds, for now. It's just that our CI machine files require some >=0.56.0 features

README.rst
```diff
@@ -1,131 +1,156 @@
...
+Obtaining Source Code
+^^^^^^^^^^^^^^^^^^^^^
 
-Latest code from git
-""""""""""""""""""""
+You can obtain the source code of the stable releases from the
+`Releases <https://github.com/taisei-project/taisei/releases>`__ page on
+GitHub, alongside the binaries. (Note that you **MUST** use the releases page,
+as the ``Download ZIP`` link on the main repo *does not* include all the code
+necessary to build the game.)
```
Akaricchi, line 113: I think it's good to say that it doesn't include submodules, specifically, to provide some context as to why this is the case to the more git savvy reader. I would say something like "as Github does not automatically include the necessary submodules in those packages", to imply that this mess isn't due to our choice.

README.rst
```diff
@@ -1,131 +1,156 @@
...
-If you cloned Taisei from git, make sure the submodules are initialized:
+You can also grab the latest code from git using the following commands:
 
 ::
 
-    git submodule init
-    git submodule update
+    git clone https://github.com/taisei-project/taisei
+    cd taisei
+    git submodule update --init --recursive
```
Akaricchi, line 121: When I said:
> Would be even better if all the info on how to properly obtain the source code could be further condensed somehow. Perhaps it's not necessary to mention the alternative way to init submodules… teaching people how git works is not a goal here.

I actually meant to just keep the `git clone` form:
```suggestion
    git clone --recurse-submodules https://github.com/taisei-project/taisei
```
i.e. `git submodule update --init` is what I referred to as the "alternative way".

README.rst
```diff
@@ -1,131 +1,156 @@
...
+You should also run ``git submodule update`` whenever you pull in
+new code, checkout another branch, or perform any ``git`` actions. The ``pull``
+and ``checkout`` helper scripts can do that for you automatically.
 
-**Important:** You should also run ``git submodule update`` whenever you pull in
-new code, checkout another branch, etc. The ``pull`` and ``checkout`` helper
-scripts can do that for you automatically.
+**Important:** Again, make sure you download the source code from either the
+``Releases`` page *or* using ``git clone``. The ``Download ZIP`` link *will not
+work!*
```
Akaricchi, line 129: I don't think it's necessary to repeat this twice. Let's trust the reader's attention a little.

README.rst
```diff
@@ -1,131 +1,156 @@
...
 
-Compiling from source
+Compiling Source Code
 ^^^^^^^^^^^^^^^^^^^^^
 
-To build and install Taisei on \*nix, just follow these steps:
+Currently, we recommend building Taisei on a \*nix or macOS-based system.
```
Akaricchi, line 134: "POSIX-like" would cover them all (and even some obscure ones like Haiku — and yes, Taisei works there without modifications; no hardware acceleration though :smile:)

README.rst
```diff
@@ -1,131 +1,156 @@
...
-    $prefix/taisei
-    $prefix/data/
+    meson setup build/
+    meson compile -C build/
+    meson install -C build/
 
-``install_relative`` is always set when building for Windows.
-
-The OpenGL ES 3.0 backend is not built by default. To enable it, do:
+You can also package the game into a ``.zip`` archive.
```
Akaricchi, line 145: I don't think it's necessary to document this here; this is a detail of the release workflow that most users won't care about. If we ever want to document releases, that can go there (along with the other packing targets e.g. tar/txz).

Akaricchi (dismissed):

README.rst
StarWitch: A bold strategy...

Akaricchi (approved):

PR #304, opened 2021-06-18
Bump `emsdk` to `2.0.25`, rework separate check for `tot` 

StarWitch:
As discussed here: https://github.com/emscripten-core/emscripten/issues/14486

Changes summary:
1. Bumps `emsdk` to `2.0.25`. Playtest confirms it works.
2. Moves the `latest` check to its own file (`emscripten-experimental.yml`) which runs at different times and is non-blocking.
3. Adds `--no-node` to defeat `closure` once and for all(?!) with this: https://github.com/emscripten-core/emscripten/pull/14501
4. Adds `-s ALLOW_UNIMPLEMENTED_SYSCALLS=1` due to this: https://github.com/emscripten-core/emscripten/pull/14491
5. Removes `-lgl` flag as it no longer works and prevents compilation.


.github/workflows/emscripten-experimental.yml
```diff
@@ -0,0 +1,88 @@
+name: 'Test Build (Emscripten Experimental)'
+on:
+  workflow_dispatch:
+  workflow_run:
```
StarWitch, line 4: `workflow_run` means it will only run after the checks in `main.yml` succeed on `master`. Prevents noise if pushes to `master` have an issue with `clang` compilation, etc.

.github/workflows/emscripten-experimental.yml
```diff
@@ -0,0 +1,88 @@
...
+    workflows: ["Test Builds"]
+    branches: [master]
+    types:
+      - completed
+      - requested
+  schedule:
```
StarWitch, line 10: If we haven't pushed to `master` in a while, `schedule` will run once a week on Friday so the `emsdk` team can see if something broke in the past week.

PR #307, opened 2021-07-03
Add free camera control tool for stage background debugging

Akaricchi:


StarWitch:
This is really cool. 

I was wondering something about this. Since the "free camera" is intended for debugging, and you mentioned being able to allocate two cameras, do you think it'd be worthwhile to have the "main camera" (i.e: the one the player is supposed to see during normal gameplay) represented by something basic, like a cube? This way, you could even see how the "main camera" moves through the stage.

Something like this ("preview camera" might be too much): 

image

Akaricchi:
> I was wondering something about this. Since the "free camera" is intended for debugging, and you mentioned being able to allocate two cameras, do you think it'd be worthwhile to have the "main camera" (i.e: the one the player is supposed to see during normal gameplay) represented by something basic, like a cube? This way, you could even see how the "main camera" moves through the stage.

I think we'd need some stronger duct tape to hold the 3D system together for something like that to be viable. This is just a quick and dirty hack. I don't think you'd want to allocate a separate camera for it, realistically. Simply disabling the animations and hooking it up to the main camera is much easier.

PR #308, opened 2021-07-04
Small fixes for the Emscripten test builders.

StarWitch:
1. Removes 'requested' as this needlessly makes it run twice.
2. Changes the automatic weekly build time to 10AM on Thursday.
3. Removes the second "ephemeral" file so there isn't a missing var in the main machine file anymore (easier local testing).

### DIFF
--- a/.github/workflows/emscripten-experimental.yml
+++ b/.github/workflows/emscripten-experimental.yml
@@ -6,10 +6,9 @@ on:
     branches: [master]
     types:
       - completed
-      - requested
   schedule:
-    # run every friday at 20:00
-    - cron: '0 20 * * 5'
+    # run every thursday at 10:00
+    - cron: '0 10 * * 4'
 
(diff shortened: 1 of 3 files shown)


Akaricchi:
> Removes the second "ephemeral" file so there isn't a missing var in the main machine file anymore (easier local testing).

I don't know about this one… The constant still doesn't have a useful value so you'd need to override it anyway, and now the only ways to do it is to either locally modify an in-tree file, or copy and still modify it. When it wasn't present in the file, you still had those two options plus a third one — to use a secondary, local machine file to specify the var. Now you just can't do that anymore. I don't see how that makes local testing any easier.

What I'd prefer is to revert this change and maybe write add a comment to the machine file, explaining that the `toolchain` constant most be supplied somehow.

StarWitch:
Done!

PR #313, opened 2021-07-10
Math/complex numbers documentation

StarWitch:
Broken out from the original PR.


doc/MATH.rst
```diff
@@ -0,0 +1,162 @@
...
+
+   // later, inside a PROJECTILE() block...
+   .move = move_asymptotic_simple(aim * offset, 5),
+
+The important piece here is the ``aim * offset`` inside the ``move()`` block.
+Being able to multiply complex numbers by each other means "procedurally"
+generating danmaku patterns becomes much easier. Multiplying two complex
+numbers together like this means their **polar coordinates** are multiplied,
+and in the case of something like ``cdir(M_PI/180 * rng_sreal())``, you can add
+"predictable randomization" to your patterns without having to wrangle with
```
laochailan, line 143: Why is this randomization predictable?

vector matricies -> matrices

-> You can quickly do rotations in your patterns without handling cumbersome matrices. In this case, we add some random scattering to the original direction of […]

doc/MATH.rst
```diff
@@ -0,0 +1,162 @@
...
+directly at the player" of ``aim`` with an additional ``offset`` angle.
+
+Additionally, the C programming language has a very robust support for handling
+complex numbers, whereas the support for things like vector matricies isn't as
+available or pleasant to use.
+
+With a bit of extra initial setup, you end up with code that's much easier to
+maintain and understand.
+
+Further Reading
```
laochailan, line 154: Maybe we can nuke the Further reading? Quaternions aren’t used in Taisei after all and I feel like we don’t need to justify why we use complex numbers. Once we have more danmaku docs, they could go here though.

doc/MATH.rst
```diff
@@ -0,0 +1,162 @@
...
+Being able to multiply complex numbers by each other means "procedurally"
+generating danmaku patterns becomes much easier. Multiplying two complex
+numbers together like this means their **polar coordinates** are multiplied,
+and in the case of something like ``cdir(M_PI/180 * rng_sreal())``, you can add
+"predictable randomization" to your patterns without having to wrangle with
+cumbersome vector matricies. You can modify the original direction of "shoot
+directly at the player" of ``aim`` with an additional ``offset`` angle.
+
+Additionally, the C programming language has a very robust support for handling
+complex numbers, whereas the support for things like vector matricies isn't as
```
laochailan, line 148: vector matricies -> vectors and matrices

laochailan (commented):
Maybe I should have made some of these comments already in the first reviewing round. Hope they help a bit.

doc/MATH.rst
```diff
@@ -0,0 +1,162 @@
...
+
+.. image:: images/math-02.png
+   :width: 300pt
+
+This also conveniently lets the enemy position ``enemy->pos`` become the new
+"origin," or ``[0, 0]``. This is useful because it means that we can more
+easily determine what angle the danmaku need to travel in to travel towards the
+player.
+
+Since we don't really care about the distance, as we're looking for an angle
```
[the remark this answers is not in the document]
StarWitch: Unsure of how to word this, as I was mostly trying to hand-wave it away as to not be too verbose. What would you recommend?

doc/MATH.rst
```diff
@@ -0,0 +1,162 @@
...
+  by Welch Labs - short and punchy video series
+* `Complex Number Fundamentals <https://www.youtube.com/watch?v=5PcpBw5Hbwo>`__
+  by 3Blue1Brown - a longer lecture-style format
+
+In-Game Examples
+''''''''''''''''
+
+Simple Movement
+^^^^^^^^^^^^^^^
+
```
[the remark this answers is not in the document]
StarWitch: Are you thinking of an entirely new section that deals with the coordinate system directly?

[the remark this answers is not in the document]
laochailan: Yes before Simple Movement

doc/MATH.rst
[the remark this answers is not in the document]
laochailan: Replacing starting from “This also conveniently lets the enemy position…” until ‘``[0, 0]`` and ``[-7,3]``”:

As a vector, ``[-4, 3]`` points from the enemy position to the player position. Its length is the distance between enemy and player. Its direction is the direction we want the danmaku to travel in. In this example, we don’t care about the distance. We want a unit length pointer towards the player. `cnormalize` does this for us by giving us a complex number with the same angle as its argument but length one.

doc/MATH.rst
```diff
@@ -0,0 +1,161 @@
...
+Being able to multiply complex numbers by each other means "procedurally"
+generating danmaku patterns becomes much easier. Multiplying two complex
+numbers together like this means adding their angles,  and in the case of
+something like ``cdir(M_PI/180 * rng_sreal())``, you can quickly do rotations in
+your patterns without handling cumbersome matrices. In this case, we add some
+random scattering to the original direction of "shoot directly at the player"
+contained in ``aim`` with an additional ``offset`` angle.
+
+Additionally, the C programming language has a very robust support for handling
+complex numbers, whereas the support for things like vectors and matricies isn't
```
laochailan, line 157: matrices typo still there

laochailan (approved):
All fine now, thanks!

doc/MATH.rst
```diff
@@ -0,0 +1,180 @@
...
+^^^^^^^^^^^^^^
+
+Of course, there'd be no point in using complex numbers over a simpler X/Y
+system if it didn't provide significant advantages downstream. So let's look at
+a more complicated danmaku pattern to see why complex numbers are more
+effective for this role.
+
+.. code-block:: c
+
+   cmplx aim = cnormalize(global.plr.pos - enemy->pos);
```
Akaricchi, line 113: Hmm, I would say `cnormalize()`, while very useful, is in itself not a good example of where complex numbers shine, because this operation is exactly analogous to its vector counterpart. The real power of complex numbers is in the multiplication semantics that were mentioned earlier (lengths are multiplied, angles are added) — that's an extremely useful property for rotations. This is acknowledged later in the example (good), so maybe that section is a better place to sing praises to complex numbers.

Overall I think we probably shouldn't try too hard to sell complex numbers as being superior to 2D vectors, just explain how we use them and highlight useful properties as well as similarities to/differences from vectors. I don't feel too strongly about this, though.

doc/MATH.rst
```diff
@@ -0,0 +1,180 @@
...
+
+This ``aim`` variable could be passed to a ``move_towards`` function attached
+to a ``PROJECTILE`` object. The effect here would be the bullets moving
+directly towards the player in a straight line, even if the player moves around
+the screen, following the player as they move around in real-time (assuming)
+more than one bullet is fired.
```
Akaricchi, line 119: I think everything after "moving directly towards the player in a straight line" only adds confusion here. "Even if the player moves around the screen, following the player as they move around in real-time" sounds like you're describing some kind of homing behavior.

doc/MATH.rst
```diff
@@ -0,0 +1,180 @@
...
+   :width: 300pt
+
+This also conveniently lets the enemy position ``enemy->pos`` become the new
+"origin," or ``[0, 0]``. This is useful because it means that we can more
+easily determine what angle the danmaku need to travel in to travel towards the
+player.
+
+As a vector, ``[-7, 3]``  points from the enemy position to the player position. Its
+length is the distance between enemy and player. Its direction is the direction
+we want the danmaku to travel in. In this example, we don’t care about the
```
Akaricchi, line 145: > Its direction is the direction we want the danmaku to travel in

It's probably better to say "angle". When we refer to a direction in code we usually mean a vector (complex) quantity, i.e. the `aim` from this example would be a "direction vector".

Akaricchi (commented):
Looks like ` doc/images/artifacts.png` was added accidentally.

Akaricchi (approved):
lgtm!

PR #319, opened 2021-08-17
Build/Maintainer Docs

StarWitch:
Split off from #311.

Good to go!

### DIFF
--- a/doc/BUILD.rst
+++ b/doc/BUILD.rst
@@ -1,5 +1,5 @@
-Taisei Project - Build/Maintainer FAQ
-=====================================
+Taisei Project - Building and Packaging
+=======================================
 
 This is intended for anyone looking to build or package Taisei Project
 for any of its supported operating systems, including Linux, macOS, Windows,
@@ -10,13 +10,43 @@ Emscripten, Switch, and others.
 Dependencies
 ------------
 
-Taisei's mandatory dependencies are as follows:
+Build-Time Dependenices
+"""""""""""""""""""""""
 
--  C (C11) compiler (``gcc``, ``clang``, etc)
--  Python >= 3.5
--  meson >= 0.56.2 (recommended)
--  ninja >= 1.10.0
--  docutils
+-  ``gcc`` or ``clang``
+-  meson >= 0.53.0 (0.56.2 recommended)
+-  Python >= 3.6
+-  `python-zstandard <https://github.com/indygreg/python-zstandard>`__ >= 0.11.1
+-  `python-docutils <https://pypi.org/project/docutils/>`__ (optional, for generating documentation)
+
+Run-Time Dependencies
+"""""""""""""""""""""
+
+Required
+''''''''
+
+-  OpenGL >= 3.3, or OpenGL ES >= 3.0
+-  SDL2 >= 2.0.10
+-  cglm >= 0.7.8
+-  libpng >= 1.5.0
+-  libwebpdecoder >= 0.5 or libwebp >= 0.5
+-  libzip >= 1.5.0 (>= 1.7.0 recommended)
+-  libzstd >= 1.4.0
+-  freetype2
+-  opusfile
+-  zlib
+
+Optional
+''''''''
+
+-  OpenSSL (for a better SHA-256 implementation; used in shader cache)
+-  SPIRV-Cross >= 2019-03-22 (for OpenGL ES backends)
+-  libshaderc (for OpenGL ES backends)
+-  `ANGLE <https://github.com/google/angle>`__ (useful for platforms with
+   flaky/non-existent OpenGL support, such as Windows)
+-  GameMode headers (Linux only; for automatic
+   `GameMode <https://github.com/FeralInteractive/gamemode>`__ integration)
+
 
 Built-In vs. System Dependencies
 """"""""""""""""""""""""""""""""
@@ -28,55 +58,36 @@ This is used to pin specific versions of libraries such as `SDL2` to known-good
 versions, as well as assist with packaging and distribution on older and/or more
 esoteric systems (i.e: Emscripten).
 
-However, recompiling every single dependency Taisei needs every time you want to
-rebuild the project is not ideal. You can still use system dependencies
-installed through your OS's package manager (`apt`, `yum`, `brew`, etc) for
-development, local testing, or simply building the game for your own system.
-
 For convenience, ``meson`` will detect which packages are missing from your
 system and use its wrap dependency system to pull in what it can. The trade-off
 is that this can result in longer build times.
 
-For releases, however, you should rely *exclusively* on the ``meson`` wrap
-dependencies with the Taisei repository, for consistency and
+For consistency, we tend to release Taisei using exclusively built-in packages.
+However, you can also use system dependencies as well. There's a tradeoff in
+consistency and reproduceability for speed and ease of use.
 
 This is controlled through the ``--wrap-mode`` flag with ``meson``. (More
 on that later.)
 
-In summary:
-
-- Development = system-install dependencies are fine (`apt`, `brew`, etc)
-- Release = ``meson`` wrap dependencies only
-
-System Dependencies
-"""""""""""""""""""
-
 Linux
 '''''
 
-(The following example is specific to Ubuntu or Debian, but you can substitute
-`apt` for the package manager for your flavour of Linux.)
-
-The following will install the mandatory tools for building Taisei.
+On an Ubuntu or Debian-based distro, he following will install the mandatory
+tools for building Taisei.
 
 .. code:: sh
 
    apt update
    apt install meson cmake build-essential
 
+Beyond that, consult the *Optional Dependencies* list above. Many distros
+package compile-time system dependencies with ``*-dev`` (i.e: ``libsdl2-dev``).
+Search with your distro's package manager to install the correct libraries.
 
-The dependencies below are technically optional, as mentioned above in
-``Built-In Vs. System Dependencies``.
-
-Keep in mind that certain packages may have different names on different
-systems. Ensure you are installing the ``dev`` versions of those packages.
-
-.. code:: sh
-
-   apt install libsdl2-dev libsdl2-mixer-dev libogg-dev libopusfile-dev libpng-dev libzip-dev libx11-dev
-
-If your distribution of Linux uses Wayland as its default window server, ensure
-that Wayland deps are installed:
+If you're using built-in dependencies *only* for whatever reason, one
+non-obvious dependency is for Wayland-based systems, where you may need to
+install the Wayland development library to ensure SDL2 compiles for it
+correctly.
 
 .. code:: sh
 
@@ -86,8 +97,7 @@ macOS
 '''''
 
 On macOS, you must install the Xcode Command Line Tools to build Taisei for
-the platform, as it contains headers and tools that aren't included in the FOSS
-versions found in ``brew``.
+the platform, as it contains headers and tools for building native macOS apps.
 
 .. code:: sh
 
@@ -101,23 +111,9 @@ tools:
 
 .. code:: sh
 
-   brew install meson cmake pkg-config docutils imagemagick pygments
+   brew install meson cmake pkg-config docutils pygments
 
-The following dependencies are technically optional, and can be pulled in at
-build-time, but they will reduce compile times during development, so it's
-recommended to install them.
-
-.. code:: sh
-
-   brew install freetype2 libzip opusfile libvorbis webp sdl2
-
-Optionally, if you're on macOS and compiling for macOS, you can to install
-`create-dmg <https://github.com/create-dmg/create-dmg>`__, which will allow
-you to have nicer-looking macOS ``.dmg`` files for distribution:
-
-.. code:: sh
-
-   brew install create-dmg
+You can then install dependencies from the *Optional Dependencies* list.
 
 As of 2021-08-05, you should **not** install the following packages via
 Homebrew, as the versions available do not compile against Taisei correctly.
@@ -125,34 +121,36 @@ If you're having mysterious errors, ensure that they're not installed.
 
 -  ``spirv-tools``
 -  ``spirv-cross``
--  ``sdl2_mixer``
 
 .. code:: sh
 
-   brew remove spirv-tools spirv-cross sdl2_mixer
+   brew remove spirv-tools spirv-cross
 
 In addition, if you're trying to compile on an older version of macOS
 (i.e: <10.12), SDL2 may not compile correctly on Homebrew (as of 2019-02-19).
 Let ``meson`` pull in the corrected version for you via subprojects.
 
-**NOTE:** While Homebrew's optional dependencies greatly improve compile times,
-if you can't remove packages that give you errors from your system for whatever
-reason, you can force ``meson`` to use its built-in subprojects by using
-``--wrap-mode`` (more on that later).
+You can also install `create-dmg <https://github.com/create-dmg/create-dmg>`__
+for packaging ``.dmg`` files, which enables some additional options such as
+positioning of icons in the ``.dmg``.
 
 Windows
 '''''''
 
 Taisei uses `mstorsjo/llvm-mingw <https://github.com/mstorsjo/llvm-mingw>`__ to
-achieve cross-compiling on Windows. Cross-compiling for Windows ends up being
-easier to maintain and more consistent than attempting to use Microsoft's native
-toolchain.
+achieve cross-compiling for Windows. Microsoft's native C compiler toolchain
+simply does not support the things Taisei needs to compile correctly, including
+fundamental things like
+`complex numbers <https://en.wikipedia.org/wiki/Complex_number>`__.
 
-On Linux, you'll need the following tools for cross-compiling Taisei for Windows
-on Linux:
 
-- ``llvm-mingw``
-- `nsis <https://nsis.sourceforge.io/Main_Page>`__ >= 3.0
+You can use ``llvm-mingw`` too, or you can check if your distro has any
+``mingw64`` cross-compiler toolchains available as well. That's is just the one
+that works for us.
+
+Additionally, you can install `nsis <https://nsis.sourceforge.io/Main_Page>`__
+(>= 3.0) for packaging Windows installer ``.exe`` files. (However, you can
+still package ``.zip`` files for Windows without it.)
 
 On macOS, you're probably better off using Docker and the
 `Docker container <https://hub.docker.com/r/mstorsjo/llvm-mingw/>`__ that
@@ -161,7 +159,9 @@ On macOS, you're probably better off using Docker and the
 Another options for Windows-based computers is leveraging Windows
 10's
 `Windows For Linux (WSL) Subsystem <https://docs.microsoft.com/en-us/windows/wsl/install-win10>`__
-to cross-compile to Windows using their Ubuntu image.
+to cross-compile to Windows using their Ubuntu image. You can also potentially
+use a ``mingw64`` toolchain directly on Windows, however that isn't supported
+or recommended, as it's generally more trouble than its worth.
 
 Checking Out Code
 -----------------
@@ -171,9 +171,7 @@ following:
 
 .. code:: sh
 
-   git clone https://github.com/taisei-project/taisei.git
-   cd taisei/
-   git submodule update --init --recursive
+   git clone --recurse-submodules https://github.com/taisei-project/taisei
 
 The ``git submodule update --init --recursive`` line is absolutely necessary,
 or Taisei will not build, as it will be missing many of the dependencies its
@@ -230,17 +228,17 @@ be covered here. Refer to the ``meson`` documentation linked above.
 Generally, ``default`` will rely on system-installed libraries when available,
 and fallback to vendored in-repository dependencies when necessary.
 
-``forcefallback`` will heavily encourage the use of in-repository dependencies
+``forcefallback`` will force the use of in-repository dependencies
 whenever possible. Recommended for release builds.
 
-``nofallback`` discourages the use of in-repository dependencies whenever
+``nofallback`` disallows the use of in-repository dependencies whenever
 possible, instead relying on system libraries. Useful for CI.
 
 .. code:: sh
 
-   # for release builds
+   # forces in-repo dependencies
    meson configure build/ --wrap-mode=forcefallback
-   # useful for testing/CI
+   # disables in-repo repositories
    meson configure build/ --wrap-mode=nofallback
 
 Relative Directory Install (``-Dinstall_relative``)
@@ -264,7 +262,8 @@ Install Prefix (``--prefix``)
 choice on your filesystem.
 
 On Linux without ``-Dinstall_relative`` enabled (i.e: ``false``), it should be
-set to ``/usr/local``.
+kept to its default ``/usr/local``. In general, don't touch it unless you need
+to.
 
 On other platforms, it will install all Taisei game files to the directory of
 your choice.
@@ -279,8 +278,8 @@ Package Data (``-Dpackage_data``)
 * Default: ``auto``
 * Options: ``auto``, ``true``, ``false``
 
-Packages data into either a glob or a ``.zip`` depending on if ``-Denable_zip``
-is ``true`` (see below).
+Packages game data into either a directory or a ``.zip`` depending on if
+``-Denable_zip`` is ``true`` (see below).
 
 .. code:: sh
 
@@ -294,6 +293,8 @@ Enable ZIP Loading (``-Denable_zip``)
 
 Controls whether or not Taisei can load game data (textures, shaders, etc) from
 ``.zip`` files. Useful for distribution and packaging.
+
+**NOTE:** Setting this to ``false`` automatically disables ``-Dpackage_data``.
 
 .. code:: sh
 
@@ -365,8 +366,8 @@ exceptions that may not be obvious to the human eye.)
 Deprecation Warnings (``-Ddeprecation_warnings``)
 """""""""""""""""""""""""""""""""""""""""""""""""
 
-* Default: ``(null)``
-* Options: ``error``, ``no-error``, ``ignore``
+* Default: ``default``
+* Options: ``error``, ``no-error``, ``ignore``, ``default``
 
 Sets deprecation warnings to either hard-fail (``error``), print as warnings but
 not trigger full errors if ``-Dwerror=true`` (``no-error``), and otherwise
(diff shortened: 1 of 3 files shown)


doc/BUILD.rst
```diff
@@ -0,0 +1,500 @@
...
+   apt-get install libsdl2-dev libsdl2-mixer-dev libogg-dev libopusfile-dev libpng-dev libzip-dev libx11-dev
+
+If your distribution of Linux uses Wayland as its default window server, ensure
+that Wayland deps are installed:
+
+.. code:: sh
+
+   apt-get install libwayland-dev
+
+For packaging, your best bet is ``.zip``. Invoke ``ninja`` to package a
```
laochailan, line 42: I don’t understand. Does `ninja zip` put the whole game into an archive? Do I want to do this as a package maintainer? Why is it my best bet?

doc/BUILD.rst
```diff
@@ -0,0 +1,500 @@
...
+
+   meson configure build/ -Dbuildtype=debug
+
+Debugging (``-Db_ndebug``)
+""""""""""""""""""""""""""
+
+* Default: ``true``
+* Options: ``true``, ``false``
+
+The name of this flag is opposite of what you'd expect. Think of it as "Not
```
laochailan, line 254: I think this affects the definition of the `NDEBUG` macro which is exclusively (?) responsible for deactivating `assert`.

doc/BUILD.rst
```diff
@@ -0,0 +1,500 @@
...
+
+   meson configure build/ -Ddeveloper=true
+
+Build Type (``-Dbuildtype``)
+""""""""""""""""""""""""""""
+
+* Default: ``release``
+* Options: ``release``, ``debug``
+
+Sets the type of build. ``debug`` enables several additional debugging features.
```
laochailan, line 241: Several additional debugging features means reduced optimizations and debug symbols right? Would be nice to clarify the difference to `-Ddeveloper`.

doc/BUILD.rst
```diff
@@ -0,0 +1,500 @@
...
+
+   /path/to/Taisei.app/Contents/MacOS/Taisei
+
+Linking & Striping (``-Db_lto``/``-Dstrip``)
+""""""""""""""""""""""""""""""""""""""""""""
+
+* Defaults: ``false``
+* Options: ``true``, ``false``
+
+These options prevent stripping of the binaries, leading to faster build times
```
laochailan, line 296: `b_lto` controls link-time optimizations. They increase build times by a lot and increases runtime performance. 
`strip=true` removes debug symbols. I never noticed that it increases build times… does it? I don’t think it makes a difference for the runtime performance. It increases binary size though.

doc/README.rst
```diff
@@ -0,0 +1,25 @@
...
+
+Graphics
+--------
+
+* `ANIMATION_FORMAT.rst <./ANIMATION_FORMAT.rst>`__ - how to produce sprite animations within the
+  game engine
+* `BASISU.rst <./BASISU.rst>`__ - how to compress certain assets within the game
+  engine
+
+Gameplay & Troubleshooting
```
laochailan, line 22: Maybe this section should go to the top.

doc/BUILD.rst
```diff
@@ -0,0 +1,500 @@
...
+.. contents::
+
+Dependencies
+------------
+
+Taisei's general dependencies are as follows:
+
+-  C (C11) compiler (``gcc``, ``clang``, etc)
+-  Python >= 3.5
+-  meson >= 0.53.0 (0.56.2 recommended)
```
laochailan, line 17: Maybe clarify `>=0.56.2` recommended.

laochailan (commented):
Thanks! Really nice to have this kind of documentation. Especially the MacOS/Windows stuff sounds like many nights of pain to get right without a manual.

I managed to spot some inaccuracies in the meson option descriptions.

doc/BUILD.rst
StarWitch: Left over from when this was a development-focused doc, mostly as a "this is how you zip it up" example. I'm going to make it clearer in the `Packaging` section later.

doc/BUILD.rst
StarWitch: 👍🏻

doc/BUILD.rst
```diff
@@ -0,0 +1,722 @@
...
+For releases, however, you should rely *exclusively* on the ``meson`` wrap
+dependencies with the Taisei repository, for consistency and
+
+This is controlled through the ``--wrap-mode`` flag with ``meson``. (More
+on that later.)
+
+In summary:
+
+- Development = system-install dependencies are fine (`apt`, `brew`, etc)
+- Release = ``meson`` wrap dependencies only
```
Akaricchi, line 49: I don't think we should prescribe a dependency management strategy for people looking to package their own distributions. In the general case, system dependencies should be preferred, if the environment provides a sane package management mechanism. The reality is that few systems actually do, which is largely the reason why we have the subprojects for official releases, but that's not the only valid way to do it. A decent Linux distro package or a macOS homebrew formula would not use them, for example.

doc/BUILD.rst
```diff
@@ -0,0 +1,722 @@
...
+.. code:: sh
+
+   apt install libsdl2-dev libsdl2-mixer-dev libogg-dev libopusfile-dev libpng-dev libzip-dev libx11-dev
+
+If your distribution of Linux uses Wayland as its default window server, ensure
+that Wayland deps are installed:
+
+.. code:: sh
+
+   apt install libwayland-dev
```
Akaricchi, line 83: Honestly I'm not sure about providing explicit package installation instructions for every distro/platform. That means maintaining dependency documentation in multiple places… Which may be challenging as demonstrated by this very example:
* freetype2 is missing
* webp/webpdecoder is missing
* libzstd is missing
* zlib is missing, though there's a very high chance it's already installed if the system has been used to compile literally anything
* docutils are missing
* sdl2-mixer is not needed
* x11 and wayland deps should not be needed, **unless** you're building SDL from a subproject, in which case you'll probably want to install a bunch more dependencies. That's a whole different can of worms; I wouldn't recommend building SDL from the subproject to anyone unless they know exactly what they're doing.

doc/BUILD.rst
```diff
@@ -0,0 +1,722 @@
...
+
+There are additional command line tools that you'll need. You can acquire those
+by using `Homebrew <https://brew.sh/>`__.
+
+Follow the instructions for installing Homebrew, and then install the following
+tools:
+
+.. code:: sh
+
+   brew install meson cmake pkg-config docutils imagemagick pygments
```
Akaricchi, line 104: I don't think imagemagick is required at build time?

doc/BUILD.rst
```diff
@@ -0,0 +1,722 @@
...
+
+The following dependencies are technically optional, and can be pulled in at
+build-time, but they will reduce compile times during development, so it's
+recommended to install them.
+
+.. code:: sh
+
+   brew install freetype2 libzip opusfile libvorbis webp sdl2
```
Akaricchi, line 112: This one isn't 100% right either:
* libvorbis isn't needed
* libpng is missing (maybe installed by default?)
* libzstd is missing

doc/BUILD.rst
```diff
@@ -0,0 +1,722 @@
...
+
+   brew install create-dmg
+
+As of 2021-08-05, you should **not** install the following packages via
+Homebrew, as the versions available do not compile against Taisei correctly.
+If you're having mysterious errors, ensure that they're not installed.
+
+-  ``spirv-tools``
+-  ``spirv-cross``
+-  ``sdl2_mixer``
```
Akaricchi, line 128: `sdl2_mixer` should have no impact as we don't depend on it anymore; as for the other two, what's the problem exactly? Maybe we can solve the issue somehow. Those shouldn't have an impact by default either though, unless you enable shader translation support. Which you'll only want in combination with GLES3/ANGLE.

doc/BUILD.rst
```diff
@@ -0,0 +1,722 @@
...
+reason, you can force ``meson`` to use its built-in subprojects by using
+``--wrap-mode`` (more on that later).
+
+Windows
+'''''''
+
+Taisei uses `mstorsjo/llvm-mingw <https://github.com/mstorsjo/llvm-mingw>`__ to
+achieve cross-compiling on Windows. Cross-compiling for Windows ends up being
+easier to maintain and more consistent than attempting to use Microsoft's native
+toolchain.
```
Akaricchi, line 149: We use that toolchain for official releases, but any mingw64-based toolchain should work fine. Some Linux distros package GCC-based mingw64 cross toolchains and many precompiled windows libraries.

Native compilation on windows is also possible with a ming64 toolchain (including that LLVM-based one that we use). But it's more complicated that it needs to be, mostly because of the symlinks in the repo. It's not "officially supported" (we do not endorse self harm) so I wouldn't expand on it in all the gory details here, but probably worth mentioning that it's possible to do if you hate yourself enough.

Also microsoft's toolchain isn't just inconvenient, their compiler just can not build Taisei. Maybe `clang-cl` can, but I never got that to work. Even I don't hate myself *that* much.

doc/BUILD.rst
```diff
@@ -0,0 +1,722 @@
...
+
+On Linux, you'll need the following tools for cross-compiling Taisei for Windows
+on Linux:
+
+- ``llvm-mingw``
+- `nsis <https://nsis.sourceforge.io/Main_Page>`__ >= 3.0
```
Akaricchi, line 155: NSIS is a very cursed and painful dependency, and you only need it if you want to create an installer. Most users probably won't need that.

doc/BUILD.rst
```diff
@@ -0,0 +1,722 @@
...
+-----------------
+
+First, you'll need to checkout the repository. You can do that with the
+following:
+
+.. code:: sh
+
+   git clone https://github.com/taisei-project/taisei.git
+   cd taisei/
+   git submodule update --init --recursive
```
Akaricchi, line 176: This can be done in one command `git clone --recurse-submodules https://github.com/taisei-project/taisei.git`

doc/BUILD.rst
```diff
@@ -0,0 +1,722 @@
...
+be covered here. Refer to the ``meson`` documentation linked above.
+
+Generally, ``default`` will rely on system-installed libraries when available,
+and fallback to vendored in-repository dependencies when necessary.
+
+``forcefallback`` will heavily encourage the use of in-repository dependencies
+whenever possible. Recommended for release builds.
+
+``nofallback`` discourages the use of in-repository dependencies whenever
+possible, instead relying on system libraries. Useful for CI.
```
Akaricchi, line 237: Nitpick, but the use of "encourage"/"discourage" is a bit confusing here, as these imply a mild suggestion instead of hard enforcement, which is what these options actually do.

doc/BUILD.rst
```diff
@@ -0,0 +1,722 @@
...
+Install Prefix (``--prefix``)
+"""""""""""""""""""""""""""""
+
+* Default: ``/usr/local``
+
+``--prefix`` installs the Taisei binary and content files to a path of your
+choice on your filesystem.
+
+On Linux without ``-Dinstall_relative`` enabled (i.e: ``false``), it should be
+set to ``/usr/local``.
```
Akaricchi, line 267: I would say that you simply should not touch `prefix` on Linux if you don't know exactly what you're doing.

doc/BUILD.rst
```diff
@@ -0,0 +1,722 @@
...
+
+   meson setup --prefix=/path/goes/here -C build/
+
+Package Data (``-Dpackage_data``)
+"""""""""""""""""""""""""""""""""
+
+* Default: ``auto``
+* Options: ``auto``, ``true``, ``false``
+
+Packages data into either a glob or a ``.zip`` depending on if ``-Denable_zip``
```
Akaricchi, line 282: What does it mean to "package data into a glob"?

doc/BUILD.rst
```diff
@@ -0,0 +1,722 @@
...
+   meson configure build/ -Dpackage_data=false
+
+Enable ZIP Loading (``-Denable_zip``)
+"""""""""""""""""""""""""""""""""""""
+
+* Default: ``true```
+* Options: ``true``, ``false``
+
+Controls whether or not Taisei can load game data (textures, shaders, etc) from
+``.zip`` files. Useful for distribution and packaging.
```
Akaricchi, line 296: Worth mentioning that setting this option to `false` also disables `package_data` by default.

doc/BUILD.rst
```diff
@@ -0,0 +1,722 @@
...
+
+Assertions (``-Db_ndebug``)
+"""""""""""""""""""""""""""
+
+* Default: ``true``
+* Options: ``true``, ``false``
+
+The name of this flag is opposite of what you'd expect. Think of it as "Not
+Debugging". It controls the ``NDEBUG`` declaration which is responsible for
+deactivating ``assert()`` functions.
```
Akaricchi, line 337: Yeah, blame the C standard for this one…

doc/BUILD.rst
```diff
@@ -0,0 +1,722 @@
...
+
+.. code:: sh
+
+   meson configure build/ -Dwerror=true
+
+Deprecation Warnings (``-Ddeprecation_warnings``)
+"""""""""""""""""""""""""""""""""""""""""""""""""
+
+* Default: ``(null)``
+* Options: ``error``, ``no-error``, ``ignore``
```
Akaricchi, line 369: There is also the `default` option, which is, surprisingly, the default value.

doc/BUILD.rst
```diff
@@ -0,0 +1,722 @@
...
+
+Link-Time Optimizations (``-Db_lto``)
+"""""""""""""""""""""""""""""""""""""
+
+* Default: ``true``
+* Options: ``true``, ``false``
+
+Link-time optimizations (LTO) increase build times, but also increase
+performance. For quicker build times during development, you can disable it.
+For release builds, this should be keep ``true``.
```
Akaricchi, line 418: typo: s/keep/kept/

doc/BUILD.rst
```diff
@@ -0,0 +1,722 @@
...
+
+Binary Striping (``-Dstrip``)
+"""""""""""""""""""""""""""""
+
+* Default: ``true``
+* Options: ``true``, ``false``
+
+This option prevents stripping of the `taisei` binary, leading to faster build
+times and keeping debugging symbols in place. There is a theoretical performance
+hit with this option enabled, but it can help with building during development.
```
Akaricchi, line 434: I don't think this affects build performance at all. This applies to the binary on installation only. Stripping is also very very fast compared to every other build step. 

Enabling this absolutely does *not* help during development, quite the opposite since you usually want to have debugging symbols in.

doc/BUILD.rst
```diff
@@ -0,0 +1,722 @@
...
+
+Note that GL ES 2.0 requires a few extensions to be present on your system
+to function correctly, most notably:
+
+- ``OES_depth_texture`` or ``GL_ANGLE_depth_texture``
+- ``OES_standard_derivatives``
+- ``OES_vertex_array_object``
+- ``EXT_frag_depth``
+- ``EXT_instanced_arrays`` or ``ANGLE_instanced_arrays`` or
+  ``NV_instanced_arrays``
```
Akaricchi, line 470: 1. We shouldn't encourage use of the gles20 backend; it is very undercooked, unmaintained, and basically useless at this point. The extensions it requires to work at all are really only available on gles3-capable hardware.
2. Probably worth noting here that shader translation is required for the gles3 backend to function properly.

doc/BUILD.rst
```diff
@@ -0,0 +1,722 @@
...
+.. code:: sh
+
+   # for GL 3.3 (default)
+   meson configure build/ -Dr_default=gl33
+   # for GL ES 3.0
+   meson configure build/ -Dr_default=gles30
+   # for GL ES 2.0
+   meson configure build/ -Dr_default=gles20
+
+The renderer can be switched in the installed ``taisei`` binary.
```
Akaricchi, line 489: What does that mean?

doc/BUILD.rst
```diff
@@ -0,0 +1,722 @@
...
+
+* Default: ``false``
+* Options: ``true``, ``false``
+
+For using ANGLE, the shader transpiler is necessary for converting Taisei's
+shaders to a format usable by that driver.
+
+.. code:: sh
+
+   meson configure build/ -Dshader_transpiler=true
```
Akaricchi, line 505: This isn't just for ANGLE; the GLES3 backend in general needs this.

doc/BUILD.rst
```diff
@@ -0,0 +1,722 @@
...
+
+* ``libEGL.(*)``
+* ``libGLESv2.(*)``
+
+The file extension can be ``.dll`` for Windows, ``.dylib`` for macOS,
+and ``.so`` for Linux.
+
+Using ``-Dinstall_angle`` and ``-Dangle_lib*`` (see above), ``meson`` will copy
+those files over into the package itself when running the packaging steps.
+
```
Akaricchi, line 571: Documenting ANGLE's cursed build process is very out of scope here, and the document is large enough as it is. The explanation of what ANGLE is and why it's used should be moved to the general `ANGLE` section, and then providing [this link](https://github.com/google/angle/blob/master/doc/DevSetup.md) is enough (with a warning that few visitors shall return in one piece and of sound mind).

doc/BUILD.rst
```diff
@@ -0,0 +1,722 @@
...
+rebuild the project is not ideal. You can still use system dependencies
+installed through your OS's package manager (`apt`, `yum`, `brew`, etc) for
+development, local testing, or simply building the game for your own system.
+
+For convenience, ``meson`` will detect which packages are missing from your
+system and use its wrap dependency system to pull in what it can. The trade-off
+is that this can result in longer build times.
+
+For releases, however, you should rely *exclusively* on the ``meson`` wrap
+dependencies with the Taisei repository, for consistency and
```
Akaricchi, line 41: and…???

Akaricchi (commented):
Sorry, took my sweet time with this again. There is a lot to cover, didn't touch the packaging and emscripten sections for now.

StarWitch:
All review comments fixed! Let me know how it looks, or when you get a chance to the build instructions.

laochailan (commented):
The documentation of the meson options looks good now!

doc/BUILD.rst
```diff
@@ -0,0 +1,730 @@
...
+
+On an Ubuntu or Debian-based distro, he following will install the mandatory
+tools for building Taisei.
+
+.. code:: sh
+
+   apt update
+   apt install meson cmake build-essential
+
+Beyond that, consult the *Optional Dependencies* list above. Many distros
```
Akaricchi, line 83: Change "Optional Dependencies" to just "Dependencies"

doc/BUILD.rst
```diff
@@ -0,0 +1,730 @@
...
+Search with your distro's package manager to install the correct libraries.
+
+If you're using built-in dependencies *only* for whatever reason, one
+non-obvious dependency is for Wayland-based systems, where you may need to
+install the Wayland development library to ensure SDL2 compiles for it
+correctly.
+
+.. code:: sh
+
+   apt install libwayland-dev
```
Akaricchi, line 94: It really goes way beyond that, e.g. if you're on X11 you need to install a bunch of X11 libs as well; if you want audio you need to install pulse or pipewire or whatever libs, etc… There's a good reason why I don't recommend building SDL as a subproject; if we really want to document it beyond "DO NOT DO THIS", it probably warrants a whole separate section.

doc/BUILD.rst
```diff
@@ -0,0 +1,730 @@
...
+``llvm-mingw`` provides, and installing ``nsis`` on top of it.
+
+Another options for Windows-based computers is leveraging Windows
+10's
+`Windows For Linux (WSL) Subsystem <https://docs.microsoft.com/en-us/windows/wsl/install-win10>`__
+to cross-compile to Windows using their Ubuntu image. You can also potentially
+use a ``mingw64`` toolchain directly on Windows, however that isn't supported
+or recommended, as it's generally more trouble than its worth.
+
+Checking Out Code
```
Akaricchi, line 166: This section should probably be first. It seems weird to have it so far down.

doc/BUILD.rst
```diff
@@ -0,0 +1,730 @@
...
+First, you'll need to checkout the repository. You can do that with the
+following:
+
+.. code:: sh
+
+   git clone --recurse-submodules https://github.com/taisei-project/taisei
+
+The ``git submodule update --init --recursive`` line is absolutely necessary,
+or Taisei will not build, as it will be missing many of the dependencies its
+needs to compile.
```
Akaricchi, line 178: Seems like you forgot to update this paragraph, since that line doesn't appear in the instructions anymore. It could be rewritten like this:

> The `--recurse-submodules` parameter is necessary. If you accidentally omitted it, you can run `cd taisei && git submodule update --init --recursive` to fix that.

doc/BUILD.rst
```diff
@@ -0,0 +1,730 @@
...
+See: `Meson Manual <https://mesonbuild.com/Wrap-dependency-system-manual.html>`__
+
+* Default: ``default``
+* Options: ``default``, ``nofallback``, ``forcefallback``, ...
+
+This is a core ``meson`` flag that does quite a few things. Not all of them will
+be covered here. Refer to the ``meson`` documentation linked above.
+
+Generally, ``default`` will rely on system-installed libraries when available,
+and fallback to vendored in-repository dependencies when necessary.
```
Akaricchi, line 229: It's more accurate to say that it'll download the missing dependencies, as we don't vendor stuff in-tree with wraps.

doc/BUILD.rst
```diff
@@ -0,0 +1,730 @@
...
+
+Assertions (``-Db_ndebug``)
+"""""""""""""""""""""""""""
+
+* Default: ``true``
+* Options: ``true``, ``false``
+
+The name of this flag is opposite of what you'd expect. Think of it as "Not
+Debugging". It controls the ``NDEBUG`` declaration which is responsible for
+deactivating ``assert()`` functions.
```
Akaricchi, line 338: Pedantic nitpick: assert is a macro, not a function

doc/BUILD.rst
```diff
@@ -0,0 +1,730 @@
...
+
+Deprecation Warnings (``-Ddeprecation_warnings``)
+"""""""""""""""""""""""""""""""""""""""""""""""""
+
+* Default: ``default``
+* Options: ``error``, ``no-error``, ``ignore``, ``default``
+
+Sets deprecation warnings to either hard-fail (``error``), print as warnings but
+not trigger full errors if ``-Dwerror=true`` (``no-error``), and otherwise
+ignore them (``ignore``).
```
Akaricchi, line 374: Probably should say that the `default` option respects the `werror` setting.

doc/BUILD.rst
```diff
@@ -0,0 +1,730 @@
...
+.. code:: sh
+
+   meson configure build/ -Ddeprecation_warnings=no-error
+
+
+Stack Trace Debugging (``-Db_sanitize``)
+""""""""""""""""""""""""""""""""""""""""
+
+This is useful for debugging crashes in the game. It uses
+`AddressSanitizer <https://github.com/google/sanitizers/wiki/AddressSanitizer>`__:
```
Akaricchi, line 387: I'd change this to something like:

```suggestion
Debugging with `Sanitizers <https://github.com/google/sanitizers/wiki>`__ (``-Db_sanitize``)
""""""""""""""""""""""""""""""""""""""""

This is useful for debugging memory management errors, leaks, and undefined behavior.
```

ASan does quite a lot more work than print stack traces.

doc/BUILD.rst
```diff
@@ -0,0 +1,730 @@
...
+
+* Default: ``false``
+* Options: ``true``, ``false``
+
+For using OpenGL ES, the shader transpiler is necessary for converting Taisei's
+shaders to a format usable by that driver.
+
+.. code:: sh
+
+   meson configure build/ -Dshader_transpiler=true
```
Akaricchi, line 511: Should be moved out of the ANGLE section.

doc/BUILD.rst
```diff
@@ -0,0 +1,730 @@
...
+libraries necessary for that engine.
+
+Generally, both need to be supplied at the same time.
+
+.. code:: sh
+
+   # for Linux
+   meson configure build/ -Dangle_libgles=/path/to/libGLESv2.dylib -Dangle_libegl=/path/to/libEGL.dylib
+   # for macOS
+   meson configure build/ -Dangle_libgles=/path/to/libGLESv2.so -Dangle_libegl=/path/to/libEGL.so
```
Akaricchi, line 529: You've got those two mixed up.

doc/BUILD.rst
```diff
@@ -0,0 +1,730 @@
...
+``.zip``, ``.tar.gz``, ``.dmg`` (for macOS), ``.exe`` (for Windows), among
+others.
+
+Examples
+""""""""
+
+Linux
+'''''
+
+Compiling on Linux for Linux is fairly straightforward. We have ``meson``
```
Akaricchi, line 596: Again, huuuuuge asterisk here because of the SDL2 nuances…

Honestly, I don't think it's worth it documenting the release/packaging options in this file. This is a fairly niche topic, and you'd really need to get into the weeds to get everything factually correct. Maybe split it off into a `RELEASE.rst` for later, and leave this one dedicated to dependencies and build options.

doc/BUILD.rst
```diff
@@ -39,7 +58,6 @@ Required
 Optional
 ''''''''
 
--  OpenSSL (for a better SHA-256 implementation; used in shader cache)
```
Akaricchi: I didn't mean to have this line removed completely. We still use sha256, and OpenSSL (or LibreSSL) can still be used to provide a better implementation. It's just that the "used in shader cache" part isn't worth mentioning.

doc/BUILD.rst
```diff
@@ -51,16 +69,16 @@ Optional
 Built-In vs. System Dependencies
 """"""""""""""""""""""""""""""""
 
-Due to the wide array of platforms Taisei supports, the project packages many of
-its core dependencies inside the project using the
+Due to the wide array of platforms Taisei supports, the project contains meson
+subprojects for its core dependencies using the
```
Akaricchi, line 73: Maybe "we provide meson subprojects" is a better way to word it.

doc/BUILD.rst
```diff
@@ -369,7 +364,7 @@ Deprecation Warnings (``-Ddeprecation_warnings``)
 * Default: ``default``
 * Options: ``error``, ``no-error``, ``ignore``, ``default``
 
-Sets deprecation warnings to either hard-fail (``error``), print as warnings but
+Sets deprecation warnings to either hard-fail (``default``/```error``), print as warnings but
```
Akaricchi, line 367: `default` and `error` aren't equivalent: `error` makes them errors regardless of whether `werror` is enabled or not, while `default` respects the `werror` setting.

doc/BUILD.rst
```diff
@@ -53,6 +53,7 @@ Required
 -  libzstd >= 1.4.0
 -  freetype2
 -  opusfile
+-  OpenSSL/LibreSSL (for a better SHA-256 implementation)
```
Akaricchi, line 56: This is an optional dependency. Other than that, everything looks fine! Sorry for taking so long to review this.

PR #320, opened 2021-08-22
lasers: improve collision; move to lasers/

Akaricchi:
Moved quantization out of the draw code. The laser is now quantized once per frame, and the segment data is used for both collisions and rendering. Player motion is now accounted for as well, preventing phasing through thin lasers. The overall result is more precise *and* faster collision detection.


src/lasers/laser.c
```diff
@@ -0,0 +1,638 @@
...
+
+cmplx las_accel(Laser *l, float t) {
+	if(t == EVENT_BIRTH) {
+		return 0;
+	}
+
+	return l->pos + l->args[0]*t + 0.5*l->args[1]*t*t;
+}
+
+cmplx las_weird_sine(Laser *l, float t) {             // [0] = velocity; [1] = sine amplitude; [2] = sine frequency; [3] = sine phase
```
laochailan, line 510: if it’s unused we don’t need it.

src/lasers/laser.c
```diff
@@ -0,0 +1,638 @@
...
+	cmplx line_normal = cimag(line_dir) - I*creal(line_dir);
+	cmplx sine_amp = l->args[1];
+	cmplx sine_freq = l->args[2];
+	cmplx sine_phase = l->args[3];
+
+	cmplx sine_ofs = line_normal * sine_amp * sin(sine_freq * t + sine_phase);
+	return l->pos + t * line_vel + sine_ofs;
+}
+
+cmplx las_sine_expanding(Laser *l, float t) { // [0] = velocity; [1] = sine amplitude; [2] = sine frequency; [3] = sine phase
```
laochailan, line 540: this one is very similar to the maxwell laser but not quite.

src/lasers/laser.c
```diff
@@ -0,0 +1,638 @@
...
+
+	return l->pos + v * t;
+}
+
+cmplx las_circle(Laser *l, float t) {
+	if(t == EVENT_BIRTH) {
+		return 0;
+	}
+
+	// XXX: should turn speed be in rad/sec or rad/frame? currently rad/sec.
```
laochailan, line 583: i would say rad/frame for consistency but doesn’t make much of a difference

laochailan (commented):
Is it possible to sample the sdf to do collisions? I guess not sensible because of the resolution dependence

Akaricchi:
@laochailan 
> Is it possible to sample the sdf to do collisions? I guess not sensible because of the resolution dependence

Theoretically possible? Yes. Good idea? No, but for a different reason. Just downloading the SDF texture from the GPU would be an order of magnitude slower than doing the math. I guess you could do collision detection on the GPU via a compute shader, but I'm not sure I'd trust that to execute deterministically… Also it's only core since GL 4.3, and not available in WebGL.

But I did the next best thing. I've ported [the distance function](https://github.com/taisei-project/taisei/blob/94c128abda676a7d7ce3d7b2782795172095bddc/src/util/geometry.c#L234) used to generate the SDF; it is now used for collision detection. It's being fed the same data too, so it should be very accurate to the visuals. 

> resolution dependence

There is no resolution dependence actually. The SDF buffer size is static (except for resolutions smaller than 800x600, because it makes no sense to have it larger than the viewport itself). One beautiful property of SDFs is that they can be scaled arbitrarily without artifacts or sharpness loss. The resolution only limits the accuracy of high frequency details.

Akaricchi:
The other code you've commented on isn't related to the PR, it's quite old actually… Unfortunately git didn't realize that `laser.c`->`lasers/laser.c` was a rename with modifications, so the diff is larger than it should've been. Sorry about that. I'll look into cleaning up those functions in a separate patch though; we can probably get rid of `las_sine_weird` at least.

PR #321, opened 2021-09-06
Machine files

StarWitch:
Machine files, split off from #289.

### DIFF
--- a/misc/ci/macos-aarch64-build-release.ini
+++ b/misc/ci/macos-aarch64-build-release.ini
@@ -1,0 +1,51 @@
+[constants]
+# apple silicon only supports MacOS 11.0, so just use that
+macos_min = '11.1'
+cflags = ['-mmacosx-version-min='+macos_min, '-arch', 'arm64']
+ldflags = cflags
+
+[host_machine]
+cpu_family = 'aarch64'
+cpu = 'aarch64'
+system = 'darwin'
+endian = 'little'
+
+[binaries]
+c = 'clang'
+cpp = 'clang++'
+objc = 'clang'
+strip = 'strip'
+pkgconfig = 'pkg-config'
+
+[properties]
+needs_exe_wrapper = false
+
+[project options]
+deprecation_warnings = 'no-error'
+release_files = true
+
+[built-in options]
+c_args = cflags
+c_link_args = ldflags
+cpp_args = cflags
+cpp_link_args = ldflags
+# necessary for -mmacosx-version-min to take effect on objc headers when building for macOS
+objc_args = cflags
+objc_link_args = ldflags
+wrap_mode = 'forcefallback'
+werror = true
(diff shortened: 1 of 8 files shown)


misc/ci/macos-aarch64-build-release.ini
```diff
@@ -0,0 +1,51 @@
+[constants]
+# apple silicon only supports MacOS 11.0, so just use that
+macos_min = '11.1'
```
Akaricchi, line 3: The comment says 11.0, but the value is 11.1?

misc/ci/macos-aarch64-build-release.ini
```diff
@@ -0,0 +1,51 @@
...
+cflags = ['-mmacosx-version-min='+macos_min, '-arch', 'arm64']
```
Akaricchi, line 4: I think for most optimal results you should also pass `-mcpu=apple-m1`. If that doesn't work (e.g. due to LLVM version not being new enough), `-mcpu=apple-a14` can be used as it [seems to be basically identical](https://github.com/llvm/llvm-project/blob/61d1cce2f83571c00f76144d42a2dea2cb3ab1ca/llvm/include/llvm/Support/AArch64TargetParser.def#L195).

misc/ci/macos-aarch64-build-release.ini
```diff
@@ -0,0 +1,51 @@
...
+ldflags = cflags
+
+[host_machine]
+cpu_family = 'aarch64'
+cpu = 'aarch64'
```
Akaricchi, line 9: Not that it really matters, but if we're passing that `-mcpu` flag, might as well set `cpu` to the same value. `cpu_family` must still be `aarch64`, though (`cpu_family` is the broad architecture/instruction set, `cpu` is the specific microarch/cpu).

misc/ci/macos-aarch64-build-release.ini
```diff
@@ -0,0 +1,51 @@
...
+
+[binaries]
+c = 'clang'
+cpp = 'clang++'
+objc = 'clang'
+strip = 'strip'
+pkgconfig = 'pkg-config'
+
+[properties]
+needs_exe_wrapper = false
```
Akaricchi, line 21: Careful with that. Might need to set this to `true` to prevent meson from running test binaries when cross-compiling (which shouldn't work, unless the build machine is also aarch64). Note that this doesn't affect native builds.

misc/ci/macos-aarch64-build-release.ini
StarWitch: `apple-m1` fails for some reason, so `apple-a14` it is!

StarWitch:
Fixed those issues with the Apple ARM file. Also removed a variable that's not used yet (since it came from #289).

PR #323, opened 2021-09-22
Stage 6 Remaster

laochailan:
All patterns are ported now. Two open ends that I can’t solve with my level of task magic are

- The framebuffer baryon effect.
- The spell practice intro doesn’t look as cool as it used to because Elly needs to spawn the baryons within the spell now. Maybe we just roll with that though.

There are also some minor comments in the code where I wasn’t sure. Probably some occurrences  of `NOT_NULL` should be considered to be replaced by `NOT_NULL_OR_DIE`. What is a good rule of thumb for that? Maybe for projectiles or enemies it is preferable to use the latter since they tend to outlive their parents.

This pull requests also contains some redesigns, most notably, newton and kepler sign.


StarWitch:
Rebased and fixed some compilation issues with `clang`, as well as making the stage end properly and on cue.

There's still a few TODOs in it but at least now it isn't full of merge conflicts!

StarWitch (approved):
I think this is ready to merge, even if it isn't perfect. You have to cut the big chunks off before you can put in the fine details, after all.

PR #326, opened 2021-10-16
Controller Documentation

StarWitch:
Mostly just moving it into its own file in `docs/`.

### DIFF
--- a/doc/CONTROLLER.rst
+++ b/doc/CONTROLLER.rst
@@ -64,7 +64,6 @@ When you're done, please consider contributing your mappings to
 and `us <https://github.com/taisei-project/SDL_GameControllerDB>`__, so
 that other people can benefit from your work.
 
-Also note that we currently only handle input from analog axes and digital
-buttons. Hats, analog buttons, and anything more exotic will not work, unless
-remapped.
-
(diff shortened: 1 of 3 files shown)


doc/CONTROLLER.rst
```diff
@@ -0,0 +1,70 @@
...
+
+When you're done, please consider contributing your mappings to
+`SDL <https://libsdl.org/>`__,
+`SDL_GameControllerDB <https://github.com/gabomdq/SDL_GameControllerDB>`__,
+and `us <https://github.com/taisei-project/SDL_GameControllerDB>`__, so
+that other people can benefit from your work.
+
+Also note that we currently only handle input from analog axes and digital
+buttons. Hats, analog buttons, and anything more exotic will not work, unless
+remapped.
```
Akaricchi, line 69: We actually do support analog trigger buttons since a while; even assigning key-press actions to analog sticks should work (though probably not useful). Situation unclear with hats/anything not present on a typical xbox controller; that stuff probably still won't work out of the box.

Akaricchi (approved):

doc/CONTROLLER.rst
StarWitch: Done!

PR #327, opened 2021-11-05
In-Game Character Profiles

StarWitch:
Adds some in-game character profiles under `Media Room -> Character Profiles`.

Preview: 
image


src/menu/charprofile.c
```diff
@@ -0,0 +1,368 @@
...
+		.fullname = "Kurumi",
+		.title = "High-Society Phlebotomist",
+		.description = "Species: Vampire\nField of Study: Phlebotomy\n\nVampiric blast from the past. Doesn’t she know Gensōkyō already has new bloodsuckers in town?\n\nSucking blood is what she’s good at, but her current job is a gatekeeper, and her true passion is high fashion.\n\nEveryone’s got multiple side-hustles these days…",
+		.unlock = "stage4boss",
+		.background = "stage4/kurumibg1",
+		.faces = {"normal", "dissatisfied", "puzzled", "tsun", "tsun_blush", "defeated"},
+	},
+	[PROFILE_IKU] = {
+		.name = "iku",
+		.fullname = "Iku",
```
Akaricchi, line 95: Nagae Iku

src/menu/charprofile.c
```diff
@@ -0,0 +1,368 @@
...
+
+	text_draw(profiles[selected].title, &(TextParams) {
+		.align = ALIGN_CENTER,
+		.pos = { 20*(1-o), 30 },
+		.shader = "text_default",
+		.color = RGBA(o, o, o, o),
+	});
+	r_mat_mv_pop();
+
+	r_color4(1, 1, 1, 0.5);
```
Akaricchi, line 251: I don't think this does anything; you're passing `color` to every draw call.

src/menu/charprofile.c
```diff
@@ -0,0 +1,368 @@
...
+		m->cursor++;
+		ctx->face = 0;
+	} else if(type == TE_MENU_ACCEPT) {
+		// show different expressions for selected character
+		ctx->face++;
+		if(!profiles[m->cursor].faces[ctx->face]) ctx->face = 0;
+	} else if(type == TE_MENU_ABORT) {
+		play_sfx_ui("hit");
+		close_menu(m);
+	}
```
Akaricchi, line 322: Since this function is currently called for every single SDL event, maybe add an `else { return false; }` clause to prevent the code below from being executed unnecessarily.

src/menu/charprofile.c
```diff
@@ -0,0 +1,368 @@
...
+static void action_show_character(MenuData *m, void *arg) {
+	return;
+}
+
+static void add_character(MenuData *m, int i) {
+	if(strcmp(profiles[i].name, "locked")) {
+		log_debug("adding character: %s", profiles[i].name);
+		portrait_preload_base_sprite(profiles[i].name, NULL, RESF_PERMANENT);
+		profiles[i].sprite = portrait_get_base_sprite(profiles[i].name, NULL);
+		MenuEntry *e = add_menu_entry(m, NULL, action_show_character, NULL);
```
Akaricchi, line 291: You could use `e->arg` to store the sprite instead of reserving space for pointers in the global array. You can also declare the profiles array `const` then.

src/menu/charprofile.c
```diff
@@ -0,0 +1,368 @@
...
+		e->transition = NULL;
+	}
+}
+
+static void charprofile_free(MenuData *m) {
+	dynarray_foreach_elem(&m->entries, MenuEntry *e, {
+		free(e->arg);
+	});
+}
```
Akaricchi, line 300: This does nothing since you aren't using `e->arg` in any of the menu entries. This would be the place to `free(m->context)`, though.

Akaricchi (commented):
Preliminary review without actually compiling/running the code yet

src/menu/charprofile.c
```diff
@@ -0,0 +1,403 @@
...
+	PROFILE_YOUMU,
+	PROFILE_CIRNO,
+	PROFILE_HINA,
+	PROFILE_SCUTTLE,
+	PROFILE_WRIGGLE,
+	PROFILE_KURUMI,
+	PROFILE_IKU,
+	PROFILE_ELLY,
+	PROFILE_LOCKED,
+} CharProfiles;
```
Akaricchi, line 41: * You don't need to define `LOCKED_PROFILE` since `PROFILE_LOCKED` exists.
* You can also specify `NUM_PROFILES` implicitly as the last entry in the `CharProfiles` enum.

src/menu/charprofile.c
```diff
@@ -0,0 +1,403 @@
...
+
+typedef struct CharacterProfile {
+	const char *name;
+	const char *fullname;
+	const char *title;
+	const char *description;
+	const char *background;
+	ProgressBGMID unlock;
+	char *faces[12];
```
Akaricchi, line 50: Should be `const char` as well.

src/menu/charprofile.h
```diff
@@ -0,0 +1,15 @@
...
+ * ---
+ * Copyright (c) 2011-2019, Lukas Weber <laochailan@web.de>.
+ * Copyright (c) 2012-2019, Andrei Alexeyev <akari@taisei-project.org>.
+ */
+
+#pragma once
+#include "taisei.h"
+
+#include "menu.h"
+#include "portrait.h"
```
Akaricchi, line 13: `#include "portrait.h"` should go into the `.c` file, because nothing in the header uses it.

src/menu/charprofile.c
```diff
@@ -0,0 +1,403 @@
...
+		.title = "Swordswoman Between Worlds",
+		.description = "Species: Half-Human, Half-Phantom\nArea of Study: Fencing\n\nThe swordswoman of the somewhat-dead.\n\n“While you were playing games, I was studying the blade” … is probably what she’d say.\n\nEntirely too serious about everything, and utterly terrified of (half of) her own existence.\n\nStill, she’s been given an important mission by Lady Yuyuko. Hopefully she doesn’t let it get to her head.",
+		.background = "youmu_bombbg1",
+		.faces = FACES("normal", "eeeeh", "embarrassed", "eyes_closed", "chuuni", "happy", "relaxed", "sigh", "smug", "surprised", "unamused"),
+	},
+	[PROFILE_CIRNO] = {
+		.name = "cirno",
+		.fullname = "Cirno",
+		.title = "Thermodynamic Ice Fairy",
+		.description = "Species: Fairy\nField of Study: Thermodynamic Systems\nThe lovably-foolish ice fairy.\n\nPerhaps she’s a bit dissatisfied after her recent duels with secret gods and hellish fairies?\n\nConsider going easy on her.",
```
laochailan, line 88: Just thermodynamics is better maybe.

src/menu/charprofile.c
```diff
@@ -0,0 +1,403 @@
...
+		.description = "Species: Insect\nField of Study: Evolutionary Socio-Entomology\n\nA bright bug - or was it insect? - far from her usual stomping grounds.\nShe feels that Insects have lost their “rightful place” in history.\nIf she thinks she has a raw evolutionary deal, someone ought to tell her about the trilobites…",
+		.unlock = PBGM_STAGE3_BOSS,
+		.background = "stage3/spellbg2",
+		.faces = FACES("normal", "calm", "outraged", "proud", "defeated"),
+	},
+	[PROFILE_KURUMI] = {
+		.name = "kurumi",
+		.fullname = "Kurumi",
+		.title = "High-Society Phlebotomist",
+		.description = "Species: Vampire\nField of Study: Phlebotomy\n\nVampiric blast from the past. Doesn’t she know Gensōkyō already has new bloodsuckers in town?\n\nSucking blood is what she’s good at, but her current job is a gatekeeper, and her true passion is high fashion.\n\nEveryone’s got multiple side-hustles these days…",
```
laochailan, line 124: To avoid repeating phlebotomist, this could be hematology, but maybe repeating is good/ok.

src/menu/charprofile.h
```diff
@@ -0,0 +1,16 @@
...
+ * Copyright (c) 2011-2019, Lukas Weber <laochailan@web.de>.
+ * Copyright (c) 2012-2019, Andrei Alexeyev <akari@taisei-project.org>.
+ */
+
+#pragma once
+#include "taisei.h"
+
+#include "menu.h"
+
+void preload_charprofile_menu(void);
```
Akaricchi, line 14: This function is not defined anywhere. I assume you were going to write it, but then added the preload calls to `add_character` instead?

I think it might be better to write this dedicated preload function anyway, and call it in both `create_charprofile_menu`  and when entering the parent "media room" menu. This way the loading delay should be less noticeable (if at all), even if the texture cache is cold and the game has to transcode some of those background textures (e.g. first-time start).

src/menu/charprofile.c
```diff
@@ -0,0 +1,409 @@
...
+			while(i) {
+				ctx->char_draw_order[i] = ctx->char_draw_order[i - 1];
+				ctx->char_draw_order[i - 1] = m->cursor;
+				--i;
+			}
+
+			break;
+		}
+	}
+}
```
Akaricchi, line 172: This seems to be copypasted from the charselect code but not actually used. In charselect the draw order matters because there we actually draw multiple characters at the same time over each other, since they have a slide-out animation (try to play around and switch between characters at different speeds in charselect to see how the animation behaves). Here we have a much simpler animation - only one character is displayed at a time and there's no slide-out; the previously selected character just disappears.

While it would be nice to implement a similar animation here for consistency, I'm not going to block the PR on it because it's a little complicated and can be added later. If it's not going to be implemented in this PR, then it's better to either remove the draw order stuff entirely, or keep it but add a TODO comment about this in the drawing code.

src/menu/charprofile.c
```diff
@@ -0,0 +1,409 @@
...
+
+	if(i != m->selected) {
+		pofs = lerp(pofs, 1, menu_fade(m));
+	}
+
+	int selected = check_unlocked_profile(m->cursor);
+
+	Color *color = RGBA(pbrightness, pbrightness, pbrightness, 1);
+	// if not unlocked, darken the sprite so it's barely visible
+	if(selected == PROFILE_LOCKED) color = RGBA(0.0, 0.0, 0.0, 0.9);
```
Akaricchi, line 243: Code style nitpick - this should be expanded into the standard block form. One-line `if`s may sometimes be acceptable when you have multiple short and similar `if` statements, but it's usually frowned upon.

Akaricchi (commented):
Three more things:

* Description formatting is a little inconsistent, some characters don't have a blank line separating the Species/Area of Study block with the rest of the text (e.g. Cirno) or between paragraphs (e.g. Scuttle).
* Description text looks a bit misaligned. It probably should fit between these imaginary lines (using the arrows as a guide).
taisei_20211221_16-58-40-740
* The menu does not have sound effects; all other menus do.

src/menu/charprofile.c
```diff
@@ -154,23 +154,6 @@ static const CharacterProfile profiles[NUM_PROFILES] = {
...
-				ctx->char_draw_order[i] = ctx->char_draw_order[i - 1];
-				ctx->char_draw_order[i - 1] = m->cursor;
-				--i;
-			}
-
-			break;
-		}
-	}
-}
-
```
Akaricchi: Should remove the draw order stuff from the context struct as well, if you're removing this.

src/menu/charprofile.c
```diff
@@ -255,6 +240,15 @@ static void charprofile_draw(MenuData *m) {
...
 		r_draw_sprite(&portrait_params);
+		r_mat_mv_push();
+		text_draw_wrapped("Press [Fire] for alternate expressions", DESCRIPTION_WIDTH, &(TextParams) {
+			.align = ALIGN_LEFT,
+			.pos = { 25, 570 },
+			.font = "standard",
+			.shader = "text_default",
+			.color = RGBA(0.9, 0.9, 0.9, 0.9),
+		});
+		r_mat_mv_pop();
```
Akaricchi, line 251: No need for `r_mat_mv_push()`/`r_mat_mv_pop()` here since you aren't modifying the matrix.

src/menu/charprofile.c
```diff
@@ -364,16 +362,17 @@ static bool charprofile_input_handler(SDL_Event *event, void *arg) {
...
-		}
-	}
 	return false;
 
 }
 
+void preload_charprofile_menu(void) {
+	for(int i = 0; i < NUM_PROFILES-1; i++) {
+		portrait_preload_base_sprite(profiles[i].name, NULL, RESF_PERMANENT);
+		preload_resource(RES_TEXTURE, profiles[i].background, RESF_OPTIONAL);
```
Akaricchi, line 372: Didn't catch this last time, but `RESF_OPTIONAL` is not appropriate here. `RESF_OPTIONAL` makes it so that the game won't crash if this preload fails, but it'll still crash when you actually try to use the failing resource, unless you acquire it through `res_texture_optional()`, e.g.
```c
Texture *tex = res_texture_optional("foobar");

if(tex) {
    use_texture(tex);
} else {
    fallback();
}
```

In 99% cases this is not what you want.

src/menu/charprofile.c
```diff
@@ -326,35 +320,39 @@ static void action_show_character(MenuData *m, void *arg) {
...
 		ctx->face = 0;
 	} else if(type == TE_MENU_CURSOR_RIGHT) {
+		play_sfx_ui("generic_shot");
 		m->cursor++;
 		ctx->face = 0;
 	} else if(type == TE_MENU_ACCEPT) {
-		// show different expressions for selected character
-		ctx->face++;
-		if(!profiles[m->cursor].faces[ctx->face]) ctx->face = 0;
+		if (check_unlocked_profile(m->cursor) != PROFILE_LOCKED) {
```
Akaricchi, line 348: No space after if

src/menu/charprofile.c
```diff
@@ -0,0 +1,395 @@
...
+		.description = "Species: Bombardier Beetle\nArea of Study: Full-Stack 'Web' Development\n\nNot all things are computable, and this bug seems to sneak in at a decisive fork in your commute to the true culprit.\n\nThe majority of insects do not scheme grand revolutions, yet their presence alone is enough to upset the course of events.",
+		.unlock = PBGM_STAGE3_BOSS,
+		.background = "stage3/spellbg1",
+		.faces = FACES("normal"),
+	},
+	[PROFILE_WRIGGLE] = {
+		.name = "wriggle",
+		.fullname = "Wriggle Nightbug",
+		.title = "Insect Rights Activist",
+		.description = "Species: Insect\nArea of Study: Evolutionary Socio-Entomology\n\nA bright bug - or was it insect? - far from her usual stomping grounds.\nShe feels that Insects have lost their “rightful place” in history.\nIf she thinks she has a raw evolutionary deal, someone ought to tell her about the trilobites…",
```
Akaricchi, line 112: Missed a paragraph separator here; the rest looks fine.

src/menu/charprofile.c
```diff
@@ -0,0 +1,395 @@
...
+	return false;
+
+}
+
+void preload_charprofile_menu(void) {
+	for(int i = 0; i < NUM_PROFILES-1; i++) {
+		portrait_preload_base_sprite(profiles[i].name, NULL, RESF_PERMANENT);
+		preload_resource(RES_TEXTURE, profiles[i].background, RESF_PERMANENT);
+	}
+};
```
Akaricchi, line 364: Actually, one more thing: the faces are not preloaded (call `portrait_preload_face_sprite` for each).

PR #332, opened 2022-01-11
build: update sdl2.wrap to 2.0.18

Akaricchi:


StarWitch:
Looks like there's a slight issue with the `darwin` section missing some files. I'll investigate after work. 

When running `meson setup ... -Dwrap_mode=forcefallback`:

```
subprojects/sdl2/src/joystick/darwin/meson.build:1:0: ERROR: File SDL_sysjoystick.c does not exist.
```

eli-schwartz:
The WrapDB had the same problem, but it was fixed: https://github.com/mesonbuild/wrapdb/pull/270

It looks like https://github.com/taisei-project/SDL/commit/1250c00db0879052dcfb7e9fc9a11b5cb7eaede3 tries to be a more feature-complete meson port (meson_options.txt is there!) so it would actually be fantastic if you could try to sync with the wrapdb version and enhance it with your changes. :D


There is a comment in your version:
```
 # Based on https://github.com/mesonbuild/sdl2/blob/2.0.3/meson.build
```

That repository is out of date, use https://github.com/mesonbuild/wrapdb/ as your base inspiration please. :)

Akaricchi:
@eli-schwartz 

Last time I looked into wrapdb, I found the patching workflow very awkward compared to just maintaining a git branch of the project with meson stuff added. That was a long time ago though, maybe something has changed?

> That repository is out of date, use https://github.com/mesonbuild/wrapdb/ as your base inspiration please. :)

Good to know, but I'm pretty sure that comment predates the new repository. The SDL wrap probably has diverged quite a lot (especially considering that I rewrote much of the original). It may be hard to integrate it back, but I'll look into it if I find the time (and if the process of going from a git repo to a patch and back is adequate…)

eli-schwartz:
The new setup should be much easier, because it actually is implemented locally as a meson project using `patch_directory` to store your meson.build patch.

You can work on `meson.build` live in subprojects/SDL2-2.0.18/ until it is in a suitable state, running the build to see if it works, and then sync the changes back into subprojects/packagefiles/sdl2/ using the command `meson subprojects packagefiles --save sdl2` (or do the reverse with `--apply`).

StarWitch:
I'm also still having problems with launching Taisei with this new branch, even after fixing the filename issue. Missing symbols and weird errors I haven't seen before. I'll send logs and investigate more tomorrow.

PR #338, opened 2022-02-12
MacOS Universal .dmg

StarWitch:
This uses a shell script to: 

* set up both build environments (for `x64` and `AArch64`)
* combine the resulting binaries together into a `combined` output directory
* create a `.dmg` with the version string in the filename (and `-universal`) using the `combined` directory

Yes, a shell script. This is much, _much_ simpler than the alternatives, which include:

* somehow getting `meson` to recognize two separate build directories, which involves adding a bunch of conditionals and other kludge for what is admittedly a very specific use-case (only macOS really requires "universal binaries" right now)
* and/or kludging up the various helper scripts in `scripts/` (a little bit of this was still necessary, but not nearly as much as the alternative of trying to mash up the two environments together)
* having every subproject understand what to do when presented with dual-arch `cflags` (which is in itself a _lot_ of work, and probably not worth it, even tho this is the "recommended" way)

Having two separate builds with two separate environments that then do a raw `lipo -create` ends up being less of a headache, and does indeed result in the combined binary you'd expect. It also means that the base Taisei meson stack doesn't really have to deal with it at all. Bonus: if this ever breaks, `ninja dmg -C build/` still works and produces single-arch `.dmg` files like before, so no change there.

You can check the output with `lipo -archs /path/to/combined/Taisei` which should come back `x86_64 arm64`. I've tested it with `x86_64` but I don't have an M1 to test `arm64` with just yet.


scripts/macos-gen-dmg.py
```diff
@@ -26,6 +26,19 @@ def main(args):
...
+        action=argparse.BooleanOptionalAction,
+    )
+
+    parser.add_argument('combine_dir',
+        help='The build directory (defaults to CWD)',
+        default=Path(os.getcwd()),
+        type=Path,
+        nargs='?',
+    )
+
```
Akaricchi, line 41: I think it makes more sense to have `combine_dir` as an argument to `--universal`, because it's meaningless without that option. Not even sure if it should be optional.

scripts/macos_build_universal.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env zsh
```
Akaricchi, line 1: Does it have to be zsh? That's probably not installed on CI by default, is it?

Would probably also need to sprinkle doublequotes around all the variable expansions to handle spaces in paths etc. Not sure if zsh needs it; bash probably does.

scripts/macos_build_universal.sh
```diff
@@ -0,0 +1,24 @@
...
+
+export BUILD_DIR=$PWD/build
```
Akaricchi, line 3: I would make this, and the Taisei git root, a command line parameter instead of hardcoding it.

scripts/macos_build_universal.sh
Akaricchi: Also probably better add `set -e` at the top of the script to stop on any error.

scripts/taiseilib/version.py
```diff
@@ -97,3 +100,7 @@ def main(args):
 
     args = parser.parse_args(args[1:])
     print(get(args=args).format(template=args.format))
+
+if __name__ == '__main__':
+    import sys
+    main(sys.argv)
```
Akaricchi, line 106: You're actually expected to call `scripts/version.py` instead of this module directly.

Akaricchi (commented):
I think the shell script solution is fine, meson is not intended to handle multi-arch builds. We probably could add a build option to build/install the binary only (and not data), to avoid doing extra compression work.

scripts/macos_build_universal.sh
StarWitch: macOS ships with `zsh` as the standard shell, and since this will (most likely, usually) be run on a Mac system I opted for `zsh`. But I also don't really use many of the fancier features in it, so I can have it set to use whatever the system shell is instead.

scripts/taiseilib/version.py
```diff
@@ -1,5 +1,8 @@
-
-from . import common
+#!/usr/bin/env python3
+try:
+    from . import common
+except:
+    import common
```
Akaricchi, line 5: Forgot to revert this.

scripts/macos_build_universal.sh
```diff
@@ -0,0 +1,32 @@
...
+export MESON_BUILD_ROOT_MACOS_X64=$BUILD_DIR/x64
+export MESON_BUILD_ROOT_MACOS_AARCH64=$BUILD_DIR/aarch64
+export MESON_BUILD_ROOT_MACOS_X64_COMPILED=$BUILD_DIR/compiled/x64
+export MESON_BUILD_ROOT_MACOS_AARCH64_COMPILED=$BUILD_DIR/compiled/aarch64
+export MESON_BUILD_ROOT_MACOS_COMBINED=$BUILD_DIR/compiled/combined
+export TAISEI_BIN_PATH=./Taisei.app/Contents/MacOS/Taisei
+
+mkdir -p $BUILD_DIR/compiled $MESON_BUILD_ROOT_MACOS_COMBINED
+meson setup --native-file $TAISEI_ROOT/misc/ci/macos-x86_64-build-release.ini --prefix $MESON_BUILD_ROOT_MACOS_X64_COMPILED $MESON_BUILD_ROOT_MACOS_X64
+meson setup --cross-file $TAISEI_ROOT/misc/ci/macos-aarch64-build-release.ini --prefix $MESON_BUILD_ROOT_MACOS_AARCH64_COMPILED $MESON_BUILD_ROOT_MACOS_AARCH64
```
Akaricchi, line 21: You are not passing `$TAISEI_ROOT` to `meson setup` here. This will not work outside of the taisei git directory.

scripts/macos-gen-dmg.py
```diff
@@ -60,4 +70,9 @@ def main(args):
 
 
 if __name__ == '__main__':
-    run_main(main)
+    try:
+        run_main(main)
+    except subprocess.CalledProcessError as e:
+        print("main dmg creation method failed, trying backup")
+        import sys
+        main(sys.argv)
```
Akaricchi, line 78: I don't understand the purpose of this. What's the "backup" and how is this supposed to work?

scripts/macos-gen-dmg.py
```diff
@@ -26,6 +26,12 @@ def main(args):
         nargs='?',
     )
 
+    parser.add_argument('universal_dmg',
+        help='If specified, creates a universal dmg (x64/arm64)',
+        type=Path,
+        nargs='?',
+    )
```
Akaricchi, line 33: This argument is pretty misleading. What it actually does (or is meant to do) is use a pre-populated installation directory instead of making a new one. It should be called something like `installation`, with the help message updated accordingly.

scripts/macos-gen-dmg.py
```diff
@@ -26,6 +26,12 @@ def main(args):
...
+
     args = parser.parse_args(args[1:])
 
     with temp_install(args.build_dir) as install_path:
```
Akaricchi, line 37: I think this should be refactored a bit. The body of the `with` block should be move into a function taking `install_path` as an argument. Then you can do something like this:

```python
if args.installation is None:
    with temp_install(args.build_dir) as install_path:
        package_dmg(install_path)
else:
    package_dmg(args.installation)
```

This way we don't do a useless `temp_install()` in case we have a premade installation.

StarWitch:
Should be good to go now.

scripts/macos-gen-dmg.py
```diff
@@ -11,6 +11,34 @@
...
+            --icon-size 64
+            --icon "Taisei.app" 100 50
+            --icon "README.txt" 50 200
+            --icon "STORY.txt" 200 200
+            --icon "GAME.html" 350 200
+            --icon "COPYING" 125 350
+            --icon "ENVIRON.html" 275 350
+            --hide-extension "Taisei.app"
+            --app-drop-link 300 50
+            {0} {1}'''.format(destination, str(install_path)))
```
Akaricchi, line 32: Splicing paths into the string is a bad idea; `shlex.split` is going to mishandle paths that contain whitespace or other significant characters like `'` and `"`. They should be appended to `command` (a list) directly instead, as they were before. In the `else` clause it is done correctly as well.

scripts/taiseilib/version.py
```diff
@@ -1,6 +1,5 @@
-
+#!/usr/bin/env python3
```
Akaricchi, line 1: You should just revert this file with `git checkout master -- scripts/taiseilib/version.py`
The shebang and +x permissions aren't needed.

PR #339, opened 2022-02-22
filewatch: rewrite inotify backend

Akaricchi:
It's now based on directory monitoring, which is more complicated, but also much more reliable, especially when dealing with move events and hard links. It is also closer to the semantics we actually want (monitoring fs paths for changes, not inodes). It will still get confused if the directory itself gets moved (directly or indirectly), though, but handling that correctly is absurdly difficult and not important for our use-case.

Importantly, `FileWatch` no longer becomes useless after reporting a `FILEWATCH_FILE_DELETED` event: if another file is placed onto the same path later, the `FileWatch` will report `FILEWATCH_FILE_UPDATED`. Some client logic can be simplified because of that.

### DIFF
--- a/src/filewatch/filewatch_inotify.c
+++ b/src/filewatch/filewatch_inotify.c
@@ -11,58 +11,238 @@
 #include "filewatch.h"
 #include "util.h"
 #include "events.h"
+#include "vfs/syspath_public.h"
 
-#include <sys/inotify.h>
-#include <unistd.h>
-#include <stdio.h>
 #include <errno.h>
+#include <stdio.h>
+#include <sys/inotify.h>
+#include <sys/resource.h>
+#include <unistd.h>
+
+// #define FW_DEBUG
+
+#ifdef FW_DEBUG
+	#undef FW_DEBUG
+	#define FW_DEBUG(...) log_debug(__VA_ARGS__)
+	#define IF_FW_DEBUG(...) __VA_ARGS__
+#else
+	#define FW_DEBUG(...) ((void)0)
+	#define IF_FW_DEBUG(...)
+#endif
+
+#define WATCH_FLAGS ( \
+	IN_CLOSE_WRITE      | \
+	IN_CREATE           | \
+	IN_DELETE           | \
+	IN_DELETE_SELF      | \
+	IN_MODIFY           | \
+	IN_MOVED_FROM       | \
+	IN_MOVED_TO         | \
+	IN_MOVE_SELF        | \
+	                      \
+	IN_EXCL_UNLINK      | \
+	IN_ONLYDIR          | \
+0)
 
 #define INVALID_WD (-1)
-
-#define WATCHED_EVENTS ( \
-	IN_CLOSE_WRITE     | \
-	IN_DELETE_SELF     | \
-	IN_MODIFY          | \
-	IN_MOVE_SELF       | \
-0)
-
-// FIXME: This doesn't detect removals of files with more than 1 hard link.
-// IM_DELETE_SELF is only emitted when the link count drops to 0.
-// Otherwise, IN_ATTRIB is emitted, indicating that the number of hard links has changed.
-// Unfortunately, IN_ATTRIB is also emitted for *any* attribute change on the file, such as
-// permissions or even access time, and there is no good way to distinguish a hard link removal from
-// anything else. In principle, we could test if the path we're interested in monitoring is still
-// accessible after an IN_ATTRIB event. However, that would require us to actually track the paths // here, complicating the code. Note that it is possible for multiple watched paths to resolve to
-// the same file/watch descriptor. In the current implementation, FileWatch instances correspond
-// 1-to-1 to watch descriptors, not paths.
-#define DELETION_EVENTS ( \
-	IN_DELETE_SELF      | \
-	IN_IGNORED          | \
-	IN_MOVE_SELF        | \
-0)
 
 #define FW (*_fw_globals)
 
 #define EVENTS_BUF_SIZE 4096
 
+typedef struct DirWatch DirWatch;
+
 struct FileWatch {
-	int wd;
-	SDL_atomic_t refs;
+	LIST_INTERFACE(FileWatch);
+	char *filename;
+	DirWatch *dw;
 	bool updated;
 	bool deleted;
 };
 
+struct DirWatch {
+	LIST_INTERFACE(DirWatch);
+	FileWatch *filewatch_list;
+	char *path;
+	int wd;
+};
+
+typedef struct WDRecord {
+	DirWatch *dirwatch_list;
+	int32_t refs;
+} WDRecord;
+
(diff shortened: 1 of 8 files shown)


src/filewatch/filewatch_inotify.c
```diff
@@ -11,59 +11,239 @@
...
+	HANDLE(IN_MOVE_SELF);
+
+	HANDLE(IN_UNMOUNT);
+	HANDLE(IN_Q_OVERFLOW);
+	HANDLE(IN_IGNORED);
+
+	HANDLE(IN_ONLYDIR);
+	HANDLE(IN_DONT_FOLLOW);
+	HANDLE(IN_EXCL_UNLINK);
+	HANDLE(IN_MASK_CREATE);
```
StarWitch, line 223: Looks like `IN_MASK_CREATE` is not available with `libnotify-kqueue` on macOS. Might need to put a check or conditional around it to see if it exists first. (Hoping this isn't load-bearing...)

See: https://github.com/libinotify-kqueue/libinotify-kqueue/blob/ea7835fcafc3cee2a0d6c0e3c8034962c48f6afe/sys/inotify.h#L81

StarWitch:
Currently, stage reloading fails with the following error on macOS:

```
9297      F: filewatch_process_events: ../src/filewatch/filewatch_inotify.c:384: assertion `((uintptr_t)_assume_aligned_ptr & ((_Alignof(struct inotify_event)) - 1)) == 0` failed
```

This is with a simple operation like removing a `INVOKE_TASK` call from a `timeline.c`.

Akaricchi:
The alignment issue smells like a bug in libinotify-kqueue. According to the Linux man page [inotify(7)](https://www.man7.org/linux/man-pages/man7/inotify.7.html) the name field should be appropriately zero-padded to preserve alignment. I've added a workaround; if that fixes the issue then we should file a bug report.

StarWitch:
Compiles without issue, and mostly-works now.

Things that do work:

* recompiling the stages 
* stage changes working upon selecting Restart Game from the in-game menu, or using `F3` with a quicksave 

Things that are still a little bit janky:

* I was editing Stage 1, mainly removing/re-adding `INVOKE_TASK_DELAYED(100, burst_fairies_1);` from the main timeline `TASK`, but it seems to recompile _all_ stages every single time, including spellcards for other stage's bosses. (i.e: I can see `src/stages/stage6/spells/toe.c` being recompiled as well), which adds a few seconds to each loop
* The watcher doesn't seem to reload at the last "quicksave" replay point set with `F4` automatically, you have to either `Restart Game` or hit `F3` to get the changes to appear

Aside from that, it did indeed fix the bug and compilation issue. Seems like a bug report is in order for the alignment issue.

Akaricchi:
> I was editing Stage 1, mainly removing/re-adding INVOKE_TASK_DELAYED(100, burst_fairies_1); from the main timeline TASK, but it seems to recompile all stages every single time, including spellcards for other stage's bosses. (i.e: I can see src/stages/stage6/spells/toe.c being recompiled as well), which adds a few seconds to each loop

Try editing the stage while the game is not running and then run `meson compile -v taisei-stages`; do that a few times. Does it actually invoke the compiler for unmodified sources? If it happens every time then it's probably a ninja or meson bug specific to macOS, though that'd be surprising. I can't reproduce it on my system. Maybe what you saw was the `src/stages/stage6/spells/toe.c.o` on the linker command line?

> The watcher doesn't seem to reload at the last "quicksave" replay point set with F4 automatically, you have to either Restart Game or hit F3 to get the changes to appear

That sounds like it doesn't detect the dynamic library file being updated/replaced, for whatever reason. To be honest, I'm losing my patience with `libinotify-kqueue`-related issues. I'll fix it if you can debug it and find out what's going on, but otherwise worksforme. You'll have to find out how your compiler/linker performs the update, which can happen in a frustrating myriad of ways. For example, on my system clang/lld writes the target into a temporary file (in the same directory), then deletes the original and moves the new file in its place, while gcc/gold yolos it and just opens the original for writing (risking a race condition if a reload is attempted before its done writing). You may want to enable `FW_DEBUG` and possibly add `IN_ALL_EVENTS` to `WATCH_FLAGS` to see all inotify events.

It's also possible that my alignment fix is just not correct and every event beyond the first in the queue is garbage. Compile with ASan and maybe add an assertion to check that `rec` is not `NULL` in `filewatch_process_event` (I can't think of any good reason why it would be, except if `e->wd` is garbage).

Alternatively, feel free to write a macOS-specific filewatch backend using Apple's blessed APIs.

StarWitch:
It was always a long shot to get it running on macOS, so I'm fine with shipping it as-is, considering I primarily work on Linux now. If I come up with some fixes in the future, I'll create another MR.

StarWitch:
Solved one mystery: it was precompiled headers causing the constant regenerate-everything behaviour. Setting `-Db_pch=false` fixes that at least.

StarWitch:
I don't understand the underlying mechanism all that thoroughly, but it seems that in `dynstage_filewatch_event()` the `FILEWATCH_FILE_DELETED` switch gets called instead of `FILEWATCH_FILE_UPDATED` for whatever reason. This might be a filesystem-level operation where the file is marked as deleted instead of overwritten, perhaps? Either way (and I know this is probably kludge-y), adding a second `dynstage_bump_lib_generation()` to the `FILEWATCH_FILE_DELETED` does cause the replay-reload functionality to work as expected on macOS.

In summary:

1. setting `-Db_pch=false` in `src/meson.build` fixes the source-regeneration issue
2. `FILEWATCH_FILE_DELETED` seems to be the event returned instead of `FILEWATCH_FILE_UPDATED`, and forcing a reload in `FILEWATCH_FILE_DELETED` fixes the replay-reload functionality

Akaricchi:
> I don't understand the underlying mechanism all that thoroughly, but it seems that in dynstage_filewatch_event() the FILEWATCH_FILE_DELETED switch gets called instead of FILEWATCH_FILE_UPDATED for whatever reason.

You should be getting a `FILEWATCH_FILE_UPDATED` almost immediately after the `FILEWATCH_FILE_DELETED`, but it gets lost for some reason. I suggest logging all inotify events and verifying the alignment fix, as I said in the previous comment. If there are no inotify events corresponding to the creation/moving-into-place of the new file, and the alignment fix actually works (i.e. `rec != NULL`), then it must be an `inotify-kqueue` bug.

> Either way (and I know this is probably kludge-y), adding a second dynstage_bump_lib_generation() to the FILEWATCH_FILE_DELETED does cause the replay-reload functionality to work as expected on macOS.

If we have to deal with missing creation events, a better way to do it (in `filewatch_inotify.c`) is to `stat()` the file shortly after a `FILEWATCH_FILE_DELETED` event, and synthesize a `FILEWATCH_FILE_UPDATED` if the file exists. Not pretty, but at least the filewatch API clients won't have to deal with this nonsense.

PR #341, opened 2022-04-17
remove recursion at vfs_mkparents_recurse

taehjoon:
FIX 'vfs_mkparents_recurse' function.
I remove recursion but can't test it in my env.

### DIFF
--- a/src/vfs/public.c
+++ b/src/vfs/public.c
@@ -10,6 +10,8 @@
 
 #include "private.h"
 
+//#include <execinfo.h>
+
 typedef struct VFSDir {
 	VFSNode *node;
 	void *opaque;
@@ -48,11 +50,16 @@ bool vfs_unmount(const char *path) {
 	return false;
 }
 
+
+//not absolute path(this path does not have zip)
 SDL_RWops* vfs_open(const char *path, VFSOpenMode mode) {
 	SDL_RWops *rwops = NULL;
 	char p[strlen(path)+1];
-	path = vfs_path_normalize(path, p);
+	
+    //This second line is similar to 'vfs_repr' in vfs/public.c
+    path = vfs_path_normalize(path, p);
 	VFSNode *node = vfs_locate(vfs_root, path);
+
 
 	if(node) {
 		assert(node->funcs != NULL);
@@ -65,6 +72,7 @@ SDL_RWops* vfs_open(const char *path, VFSOpenMode mode) {
 	} else {
 		vfs_set_error("Node '%s' does not exist", path);
 	}
+    log_debug("VFS_OPEN_HAS_ERROR?");
 
 	return rwops;
 }
@@ -124,7 +132,7 @@ void vfs_mkdir_required(const char *path) {
 		log_fatal("%s", vfs_get_error());
 	}
 }
-
+/*
 static bool vfs_mkparents_recurse(const char *path) {
 	// FIXME this has stupid space complexity and is probably silly in general; optimize if you care
 
@@ -137,7 +145,8 @@ static bool vfs_mkparents_recurse(const char *path) {
 
 	char p[strlen(path) + 1];
 	strcpy(p, path);
-	psep += p - path;
+	log_debug("MKP_REC %p %p",p ,path);
+    psep += p - path;
 	*psep = 0;
 
 	VFSInfo i = vfs_query(p);
@@ -152,6 +161,36 @@ static bool vfs_mkparents_recurse(const char *path) {
 
 	return vfs_mkparents_recurse(p) && vfs_mkdir(p);
 }
+*/
+static bool vfs_mkparents_recurse(const char *path) {
+
+    char p[strlen(path) + 1];
+    VFSInfo i;
+    strcpy(p,path);
+    while(*p){
+        char *psep = strrchr(p, VFS_PATH_SEPARATOR);
+
+        if(!psep){
+            //current parent is root
+            return true;
+        }
+
+        *psep = 0;
+
+        i = vfs_query(p);
+
+        if(i.error){
+            return false;
+        }else if(i.exists){
+            return true;
+        }else if(!vfs_mkdir(p)){
+            return true;
+        }
+    }
+    
+    return false;
+}
+
 
 bool vfs_mkparents(const char *path) {
 	char p[strlen(path)+1];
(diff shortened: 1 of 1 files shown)


StarWitch:
Thank you for the PR. It'd be good if you could edit your description tell us what problem you're trying to solve with this.

I'll give it a general review for now.

src/vfs/public.c
```diff
@@ -10,6 +10,8 @@
 
 #include "private.h"
 
+//#include <execinfo.h>
```
StarWitch, line 13: Remove commented out imports.

src/vfs/public.c
```diff
@@ -48,12 +50,17 @@ bool vfs_unmount(const char *path) {
 	return false;
 }
 
+
+//not absolute path(this path does not have zip)
```
StarWitch, line 54: Either elaborate what you mean by this, or delete the comment.

src/vfs/public.c
```diff
@@ -48,12 +50,17 @@ bool vfs_unmount(const char *path) {
...
 SDL_RWops* vfs_open(const char *path, VFSOpenMode mode) {
 	SDL_RWops *rwops = NULL;
 	char p[strlen(path)+1];
-	path = vfs_path_normalize(path, p);
+	
+    //This second line is similar to 'vfs_repr' in vfs/public.c
+    path = vfs_path_normalize(path, p);
 	VFSNode *node = vfs_locate(vfs_root, path);
 
+
```
StarWitch, line 63: Remove extra newline.

src/vfs/public.c
```diff
@@ -65,6 +72,7 @@ SDL_RWops* vfs_open(const char *path, VFSOpenMode mode) {
 	} else {
 		vfs_set_error("Node '%s' does not exist", path);
 	}
+    log_debug("VFS_OPEN_HAS_ERROR?");
```
StarWitch, line 75: Either make this debug message more explicit and useful, or remove it.

src/vfs/public.c
```diff
@@ -124,7 +132,7 @@ void vfs_mkdir_required(const char *path) {
 		log_fatal("%s", vfs_get_error());
 	}
 }
-
+/*
```
StarWitch, line 135: If you intended to comment out/rewrite this entire function, simply delete the whole thing.

src/vfs/public.c
```diff
@@ -160,10 +199,28 @@ bool vfs_mkparents(const char *path) {
 }
 
 char* vfs_repr(const char *path, bool try_syspath) {
-	char buf[strlen(path)+1];
-	path = vfs_path_normalize(path, buf);
-	VFSNode *node = vfs_locate(vfs_root, path);
+    char buf[strlen(path)+1];
+/*
```
StarWitch, line 203: Again, large comment blocks of dead code should be deleted outright. No point keeping them around.

taehjoon:
thanks for your advices, I will repair it appropriately ASAP.

taehjoon:
but, I just remove recursion in 'vfs_mkparents_recurse' because this function has lots of memories.
If didn't need it, I will remove it.

src/vfs/public.c
```diff
@@ -48,12 +50,17 @@ bool vfs_unmount(const char *path) {
...
 
+
+//not absolute path(this path does not have zip)
 SDL_RWops* vfs_open(const char *path, VFSOpenMode mode) {
 	SDL_RWops *rwops = NULL;
 	char p[strlen(path)+1];
-	path = vfs_path_normalize(path, p);
+	
+    //This second line is similar to 'vfs_repr' in vfs/public.c
+    path = vfs_path_normalize(path, p);
```
Akaricchi, line 60: This comment is misleading; `vfs_repr` and `vfs_path_normalize` do fundamentally different things. 

`vfs_repr` returns a human-readable string with information about a VFS path suitable for printing out, e.g. a "real" system path that the VFS node corresponds to, or a path that points inside of a zip-file. Its output is intended for diagnostics only.

`vfs_path_normalize` takes a VFS path and converts it into its canonical form. The result is always a VFS path that points to the same object as the original path. For example, it will convert `/foo///bar/../example` and `foo/./example` both to `foo/example`. Unfortunately, the guarantees of this canonical form are not yet documented.

src/vfs/public.c
```diff
@@ -48,12 +50,17 @@ bool vfs_unmount(const char *path) {
 	return false;
 }
 
+
+//not absolute path(this path does not have zip)
 SDL_RWops* vfs_open(const char *path, VFSOpenMode mode) {
```
Akaricchi, line 55: It's more correct to say that this function takes virtual paths, i.e. VFS paths, as opposed to real filesystem paths. "Relative" is confusing, as it means something else, and is only applicable to real filesystem paths (VFS paths are always "absolute", since there is no "working directory" equivalent in the VFS).

src/vfs/public.c
```diff
@@ -152,6 +161,36 @@ static bool vfs_mkparents_recurse(const char *path) {
 
 	return vfs_mkparents_recurse(p) && vfs_mkdir(p);
 }
+*/
+static bool vfs_mkparents_recurse(const char *path) {
+
+    char p[strlen(path) + 1];
+    VFSInfo i;
+    strcpy(p,path);
```
Akaricchi, line 169: Since the length of the string is known, just use `memcpy`:

```suggestion
    memcpy(p, path, sizeof(p));
```

src/vfs/public.c
```diff
@@ -160,10 +199,28 @@ bool vfs_mkparents(const char *path) {
 }
```
Akaricchi, line 199: Since `vfs_mkparents_recurse` no longer actually recurses, it makes sense to just merge it back into `vfs_mkparents`. This way you also avoid making a redundant copy of the string.

src/vfs/public.c
```diff
@@ -152,6 +161,36 @@ static bool vfs_mkparents_recurse(const char *path) {
...
+        }
+
+        *psep = 0;
+
+        i = vfs_query(p);
+
+        if(i.error){
+            return false;
+        }else if(i.exists){
+            return true;
```
Akaricchi, line 185: If it already exists, then we should only return `true` if it's a directory (`i.is_dir`); otherwise that's an error condition (return false).

src/vfs/public.c
```diff
@@ -152,6 +161,36 @@ static bool vfs_mkparents_recurse(const char *path) {
...
+        }else if(!vfs_mkdir(p)){
+            return true;
```
Akaricchi, line 187: The logic here seems wrong as well. `vfs_mkdir` returns `true` on success, in which case we're done. If it fails, though, that's when we have to keep going and repeat the process for the parents — **and then retry `mkdir`ing the original path**. That part in bold is likely the main reason why the original function is recursive — the algorithm requires a stack. To remove the recursion I'd think you have to iterate forwards instead, but whether that is a win is not clear. Such an algorithm would have the worst-case performance in the "no-op" case (when all parents already exist).

src/vfs/public.c
```diff
@@ -160,10 +199,28 @@ bool vfs_mkparents(const char *path) {
...
+    int nptrs = backtrace(buffer, 100);
+    char ** strings;
+    strings = backtrace_symbols(buffer,nptrs);
+    if(strings != NULL){
+        //functions are 0~nptrs
+        log_debug("functions list");
+        for(int j=0; j<nptrs; j++){
+            log_debug("%s",strings[j]);
+        }    
+    }
```
Akaricchi, line 214: I recommend compiling with `meson configure -Db_sanitize=address,undefined` for debugging; you won't have to resort to hacks like these in 99% of cases.

taehjoon:
I want to make this function more efficient, but it seems to work differently than I thought. I've been having trouble verifying the code for other things these days, and I'm closing this request because I think it needs more revisions.

PR #358, opened 2023-02-23
fix errors found by muon analyze

annacrombie:
I found a few small errors while testing muon analyze on this project.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -544,12 +544,15 @@ config.set('TAISEI_BUILDCONF_TESTING_STAGES', use_testing_stages)
 
 # Stolen from Sway
 # Compute the relative path used by compiler invocations.
-source_root = meson.current_source_dir().split('/')
-build_root = meson.project_build_root().split('/')
+source_root = meson.current_source_dir()
+build_root = meson.project_build_root()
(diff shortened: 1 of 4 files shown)


meson.build
```diff
@@ -544,12 +544,15 @@ config.set('TAISEI_BUILDCONF_TESTING_STAGES', use_testing_stages)
 
 # Stolen from Sway
 # Compute the relative path used by compiler invocations.
-source_root = meson.current_source_dir().split('/')
-build_root = meson.project_build_root().split('/')
+source_root = meson.current_source_dir()
+build_root = meson.project_build_root()
 if build_machine.system() == 'windows'
     source_root = source_root.replace('\\', '/')
```
annacrombie, line 550: no .replace method on list.

Akaricchi:
Thanks. Can muon actually build this yet?

annacrombie:
Yes muon can build taisei with a small patch to add `cpp` as a language in the main project() invocation.  This issue is likely hidden with meson since in meson there is a global dict for  compilers across all subprojects, so if a subproject uses `cpp` then it is available in the main project.

PR #359, opened 2023-02-24
Remove old pre-task-system code

Akaricchi:
Now that all stages have been finally ported, we have quite a bit of dead code to get rid of.


StarWitch:
image

It's so beautiful.

StarWitch:
Currently, the only deprecation warnings I get while compiling have to do with `move_towards` and its wrapper.

```
./src/stages/stage6/nonspells/boss_nonspell_6.c:135:15: warning: '_move_towards_nargs2' is deprecated: Use move_towards(velocity, target, attraction) or move_from_towards(origin, target, attraction) instead [-Wdeprecated-declarations]
```

Some other random things:

* `src/util/glm.h` -- `glm_ease_apply` appears unused
* `STAGE3D_DEFAULT_ASPECT` is only used in `stage1` and might be easily removable

src/credits.c
```diff
@@ -520,16 +520,15 @@ static void credits_finish(void *arg) {
 }
 
 static void credits_process(void) {
```
laochailan, line 522: probably this could be a task too?

src/enemy.c
```diff
@@ -364,7 +309,7 @@ float enemy_get_hurt_radius(Enemy *enemy) {
 
 static bool should_auto_kill(Enemy *enemy) {
 	return
-		(enemy->hp <= 0 && enemy->hp != _internal_ENEMY_IMMUNE) ||
```
laochailan, line 367: why is this no longer needed?

src/lasers/laser.c
```diff
@@ -82,14 +75,20 @@ Laser *create_laserline(cmplx pos, cmplx dir, float charge, float dur, const Col
...
+		.laser = ENT_BOX(l),
+		.charge_delay = charge,
+		.target_width = width,
+	);
+	return l;
 }
 
 void laserline_set_ab(Laser *l, cmplx a, cmplx b) {
 	l->pos = a;
-	l->args[0] = (b - a) * 0.005;
```
laochailan, line 92: wait what was going on here?

laochailan (approved):

StarWitch (approved):

src/credits.c
Akaricchi: Probably, but that can be refactored later. Credits code isn't great in general. I took the path of least resistance here for now.

src/enemy.c
Akaricchi: It's not used anymore. The same effect can be achieved by adding `EFLAG_NO_AUTOKILL` to `enemy->flags`. `EFLAGS_GHOST` is a set of flags that generally matches the old behavior of `ENEMY_IMMUNE`.

src/lasers/laser.c
Akaricchi: See `create_laserline_ab`:

```c
Laser *create_laserline_ab(cmplx a, cmplx b, float width, float charge, float dur, const Color *clr) {
	float lt = 200;
	cmplx m = (b - a) / lt;
	Laser *l = create_laser(a, 200, dur, clr, las_linear, m, 0, 0, 0);
	INVOKE_TASK(laser_charge,
		.laser = ENT_BOX(l),
		.charge_delay = charge,
		.target_width = width,
	);
	return l;
}
```

200 was chosen somewhat arbitrarily (by you) for the lifetime parameter here a long time ago, and 0.005 is the reciprocal of it. I made the `laserline_set_ab` function a little more general by taking the actual lifetime into account rather than hardcoding it.

Also, now that you brought it to my attention, I noticed that I forgot to replace one instance of 200 with lt in `create_laserline_ab`.

PR #373, opened 2023-10-10
switch: use native screen resolution and remove old workarounds

p-sam:
I had to apply the same kind of workaround needed for the meson bug encountered with libzstd. Most Mesa and libnx bugs we encountered have since been fixed, so I tried removing quirks and ifdef'd code in the switch code. 

Also, the SDL window size will now adjust to the native resolution of the screen you're playing on (tv, handheld). This is mostly handheld by the switch SDL port under the hood, which resizes the window if not flagged fullscreen.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -499,6 +499,9 @@ void config_load(void) {
 #ifdef __SWITCH__
 	config_set_int(CONFIG_GAMEPAD_ENABLED, true);
 	config_set_str(CONFIG_GAMEPAD_DEVICE, "any");
+	config_set_int(CONFIG_FULLSCREEN, 0);
+	config_set_int(CONFIG_VID_WIDTH, nxGetInitialScreenWidth());
+	config_set_int(CONFIG_VID_HEIGHT, nxGetInitialScreenHeight());
 #endif
 }
 
--- a/src/config.c
+++ b/src/config.c
@@ -499,7 +499,6 @@ void config_load(void) {
 #ifdef __SWITCH__
 	config_set_int(CONFIG_GAMEPAD_ENABLED, true);
 	config_set_str(CONFIG_GAMEPAD_DEVICE, "any");
-	config_set_int(CONFIG_FULLSCREEN, 0);
 	config_set_int(CONFIG_VID_WIDTH, nxGetInitialScreenWidth());
 	config_set_int(CONFIG_VID_HEIGHT, nxGetInitialScreenHeight());
 #endif
--- a/src/video.c
+++ b/src/video.c
@@ -142,6 +142,8 @@ static VideoCapabilityState video_query_capability_alwaysfullscreen(VideoCapabil
 
 static VideoCapabilityState video_query_capability_switch(VideoCapability cap) {
 	switch(cap) {
+		// We want the window to be resizable and resized internally by SDL
+		// when the Switch gets docked/undocked
 		case VIDEO_CAP_FULLSCREEN:
 			return VIDEO_NEVER_AVAILABLE;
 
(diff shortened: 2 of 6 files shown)


src/config.c
```diff
@@ -499,6 +499,9 @@ void config_load(void) {
 #ifdef __SWITCH__
 	config_set_int(CONFIG_GAMEPAD_ENABLED, true);
 	config_set_str(CONFIG_GAMEPAD_DEVICE, "any");
+	config_set_int(CONFIG_FULLSCREEN, 0);
+	config_set_int(CONFIG_VID_WIDTH, nxGetInitialScreenWidth());
+	config_set_int(CONFIG_VID_HEIGHT, nxGetInitialScreenHeight());
```
Akaricchi, line 504: Not sure what this is for. If SDL is supposed to automatically resize the window based on the docked/undocked state, it should enforce the correct size when the window is created too. Does it not do that?

meson.build
```diff
@@ -208,7 +208,7 @@ dep_webp        = dependency('libwebp',        version : '>=0.5',   required : t
 dep_webpdecoder = dependency('libwebpdecoder', version : '>=0.5',   required : false)
 dep_zlib        = dependency('zlib',                                required : true)
 dep_zstd        = dependency('libzstd',        version : '>=1.4.0', fallback : ['libzstd', 'libzstd_dep'])
-dep_zip         = dependency('libzip',         version : '>=1.5.0', required : opt_vfs_zip, allow_fallback : true)
+dep_zip         = dependency('libzip',         version : '>=1.5.0', required : opt_vfs_zip, fallback : ['libzip', 'libzip_dep'])
```
Akaricchi, line 211: Can you explain what was the problem here and/or do you have a failed build log?

I think the problem with zstd was related to `meson.override_dependency()` misbehaving when the dependency is needed for both native and cross targets, but I don't remember the details. I wouldn't think it's the same problem here; IIRC libzip is only used by taisei itself.

src/config.c
p-sam: Yeah it doesn't, because Switch SDL port only handles the resize after the proper event, and not initially, but that's probably irrelevant since the width/height present in the config is used to create the window anyway, which would be whatever resolution was the native when Taisei was last launched

src/video.c
```diff
@@ -757,6 +775,11 @@ static bool video_handle_window_event(SDL_Event *event, void *arg) {
 			// It's followed by SDL_WINDOWEVENT_RESIZED for external resizes (from the WM or the
 			// user). We only need to handle external resizes.
 			log_debug("SDL_WINDOWEVENT_SIZE_CHANGED: %ix%i", event->window.data1, event->window.data2);
+
+			// Catch resizes by the SDL portlib itself, when the console is docked/undocked
+			if(video_get_backend() == VIDEO_BACKEND_SWITCH) {
+				video_handle_resize(event->window.data1, event->window.data2);
+			}
```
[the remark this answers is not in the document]
p-sam: Sure, I now have an issue on their fork for that, and will let you know if anything happens

meson.build
p-sam: The relevant error line in the build log was: 
```
WARNING: Subproject 'libzip' did not override 'libzip' dependency and no variable name specified
```
I eventually traced back the error to an [issue you posted on the meson repo](https://github.com/mesonbuild/meson/issues/11189) which led me to believe this was the same problem.

Akaricchi: Was that built in a pristine container or just locally? In the later case you might have an older version of `libzip` checked out under `subprojects`, which doesn't call `meson.override_dependency()`. See if it still happens after a `meson subprojects update` or just `rm -rf subprojects/libzip`.

But if this happens under a totally clean checkout, that is very odd and needs further investigation.

p-sam: It does not happen after a clean checkout, so I removed the change from the PR

src/config.c
Akaricchi: This falls under the same category as the `SDL_WINDOWEVENT_SIZE_CHANGED` thing for me… I'm pretty unhappy about this strange implementation choice in the SDL port and would consider it a bug, but I'm ok with keeping this workaround until it's fixed, if ever.

Akaricchi:
Massaged this a little bit and merged manually into `staging`

PR #389, opened 2024-10-14
Bump macOS GHA builds to `macos-15`, default to ARM64.

StarWitch:
GHA macOS runners are now ARM64-default, and this required a few changes to make it work.

We're almost at the release of the M4 processor, which means the Mac platform is now ~4 years into the move to ARM64, which means we need to start assuming this as the default.

I also adjusted the standard CI test to represent aarch64 as the default.

Releases have also been tested: [link](https://github.com/StarWitch/taisei/actions/runs/11333909194/job/31518999168)


.github/workflows/main.yml
```diff
@@ -29,7 +29,7 @@ on:
 env:
   MESON_VERSION: '0.63.3'
   EM_VERSION: '3.1.68'
-  PYTHON_VERSION: '3.11.8'
+  PYTHON_VERSION: '3.11'
```
Akaricchi, line 32: What's up with this change?

misc/ci/macos-aarch64-build-test-ci.ini
```diff
@@ -1,13 +1,9 @@
 [constants]
-macos_min = '10.15'
-cflags = ['-mmacosx-version-min='+macos_min, '-arch', 'x86_64']
+macos_min = '11.1'
```
Akaricchi, line 2: We can probably just bump this all the way to 13.0. That's the oldest version still supported by Apple.

Akaricchi (commented):
lgtm % comments

.github/workflows/main.yml
StarWitch: The images we use currently provide the latest `3.11.x` dot-releases, which is something like `3.11.10`.
 
Basically, by specifying `3.11.8`, what would happen is that later revisions of `3.11.x` would be uninstalled, at this older dot-release was being installed. It would uninstall `3.11.10` and install `3.11.8`, which seemed really silly.

tl;dr: This speeds up the build process by about 10-15 seconds.

PR #393, opened 2024-11-22
util/consideredharmful: Use overloadable func attribute with clang

kraj:
When building with glibc HEAD, it has fortified headers with clang as well and clang reports errors e.g.

| ../git/src/util/consideredharmful.h:33:7: error: redeclaration of 'strncat' must have the 'overloadable' attribute
|    33 | char* strncat();
|       |       ^
| /mnt/b/yoe/master/build/tmp/work/core2-64-yoe-linux/taisei/1.4.2/recipe-sysroot/usr/include/bits/string_fortified.h:145:8: note: previous overload of function is here
|   145 | __NTH (strncat (__fortify_clang_overload_arg (char *, __restrict, __dest),
|       |        ^


Akaricchi:
Hi, I'm sorry for the extremely late response. Overloading is not really correct here, as that would be declaring a new symbol. This feature is also not available in GCC AFAIK. I've come up with an alternative solution that should be more portable. Please check 1e8a2d916 and see if that patch fixes your issue.

PR #404, opened 2025-08-24
The Chinese Edition of STORY.txt

Swordsdeer:
As the title suggests, I've roughly translated the STORY.txt. There are many Chinese Touhou players following your game (myself included), so translating the storyline is beneficial. Hope it can be approved.


laochailan:
辛苦了，谢谢！
If you feel up to it, a translation of GAME.rst is also welcome.

PR #410, opened 2025-08-30
update zh_CN.po

Swordsdeer:
Hi, I've translated a portion of the content. (Actually, my local progress is a bit ahead, but I still haven’t managed to build a working binary for testing, so I’m hesitant to upload it. Besides, with school starting soon, time is tight. I’m afraid it might be difficult to complete the localization before 2026.)


laochailan:
Also for the credits, is it okay to include your email address from your github profile or do you want another handle/handles there?

PR #422, opened 2026-02-06
emmmmm

Swordsdeer:
What Can I Say


onihilist:
@Swordsdeer

laochailan:
Thanks for the work! Thank you to Renko too!

@onihilist there have been about 10 short new strings. When you have time, you can add them to the french too!

PR #435, opened 2026-07-25
switch: port fixes

p-sam:
* fix switch port compilation with recent commits
* updated Dockerfile

**Note:** requires https://github.com/taisei-project/SDL/pull/3 to be merged

### DIFF
--- a/misc/ci/Dockerfile.switch
+++ b/misc/ci/Dockerfile.switch
@@ -1,4 +1,4 @@
-FROM debian:bullseye-slim
+FROM debian:bookworm-slim
 
 LABEL maintainer="Alice D. <alice@starwitch.productions>"
 
@@ -19,22 +19,23 @@ RUN apt update && \
 		git \
 		libarchive-tools \
 		doxygen \
+		docutils-common \
+		gettext \
 		gnupg \
 		build-essential \
 		python3-docutils \
 		python3-pip \
+		python3-gnupg \
 		gdebi-core \
 		apt-transport-https \
-		cmake
+		cmake \
+		ninja-build
 
-RUN pip3 install \
-		meson==0.63.3 \
-		ninja \
-		zstandard \
-		python-gnupg
+RUN pip3 install --break-system-packages meson backports.zstd 
 
 ENV DEVKITPRO=/opt/devkitpro
 ENV PATH=${DEVKITPRO}/tools/bin:$PATH
+ENV CMAKE=/usr/bin/cmake
 
 RUN mkdir -p /usr/share/keyring/ && \
 	wget -U "dkp apt" -O /usr/share/keyring/devkitpro-pub.gpg https://apt.devkitpro.org/devkitpro-pub.gpg && \
@@ -58,8 +59,7 @@ RUN apt-get install -y --no-install-recommends gh
 
 RUN apt-get clean -y
 
-# required so that dkp-pacman works, otherwise it exits with an error because /etc/mtab doesn't exist
-RUN ln -s /proc/self/mounts /etc/mtab
+RUN [ -e /etc/mtab ] || ln -s /proc/self/mounts /etc/mtab
 
 RUN dkp-pacman -Syyu --noconfirm \
 		switch-dev \
--- a/src/vfs/setup_switch.c
+++ b/src/vfs/setup_switch.c
@@ -9,25 +9,26 @@
 
 #include "setup.h"
 
-#include "util/stringops.h"
+#include "memory/scratch.h"
+#include "util/strbuf.h"
 
 static void vfs_setup_onsync(CallChainResult ccr) {
 	const char *program_dir = nxGetProgramDir();
-	char *res_path = strfmt("%s/%s", program_dir, TAISEI_BUILDCONF_DATA_PATH);
-	char *storage_path = strfmt("%s/storage", program_dir);
-	char *cache_path = strfmt("%s/cache", program_dir);
+	StringBuffer buf = { acquire_scratch_arena() };
+	VfsSetupFixedPaths paths = {};
 
-	VfsSetupFixedPaths paths = {
-		.res_path = res_path,
-		.storage_path = storage_path,
-		.cache_path = cache_path,
-	};
+	strbuf_printf(&buf, "%s/%s", program_dir, TAISEI_BUILDCONF_DATA_PATH);
+	paths.res_path = strbuf_commit(&buf);
+
+	strbuf_printf(&buf, "%s/storage", program_dir);
+	paths.storage_path = strbuf_commit(&buf);
+
+	strbuf_printf(&buf, "%s/cache", program_dir);
+	paths.cache_path = strbuf_commit(&buf);
 
 	vfs_setup_fixedpaths(&paths);
 
-	free(res_path);
-	free(storage_path);
-	free(cache_path);
+	release_scratch_arena(buf.arena);
 
 	vfs_setup_onsync_done(ccr);
 }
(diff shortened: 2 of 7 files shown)


src/vfs/setup_switch.c
```diff
@@ -9,25 +9,30 @@
...
-	char *res_path = strfmt("%s/%s", program_dir, TAISEI_BUILDCONF_DATA_PATH);
-	char *storage_path = strfmt("%s/storage", program_dir);
-	char *cache_path = strfmt("%s/cache", program_dir);
+	StringBuffer res_path_buf = { acquire_scratch_arena() };
+	StringBuffer storage_path_buf = { acquire_scratch_arena() };
+	StringBuffer cache_path_buf = { acquire_scratch_arena() };
+
+	strbuf_printf(&res_path_buf, "%s/%s", program_dir, TAISEI_BUILDCONF_DATA_PATH);
+	strbuf_printf(&storage_path_buf, "%s/storage", program_dir);
+	strbuf_printf(&cache_path_buf, "%s/cache", program_dir);
```
Akaricchi: You only need one

```c
StringBuffer buf = { acquire_scratch_arena() };
VfsSetupFixedPaths paths = {};

strbuf_printf(&buf, "%s/%s", program_dir, TAISEI_BUILDCONF_DATA_PATH);
paths.res_path = strbuf_commit(&buf);

strbuf_printf(&buf, "%s/storage", program_dir);
paths.storage_path = strbuf_commit(&buf);

strbuf_printf(&buf, "%s/cache", program_dir);
paths.cache_path = strbuf_commit(&buf);

vfs_setup_fixedpaths(&paths);

release_scratch_arena(buf.arena);
```

src/vfs/syspath_posix.c
```diff
@@ -17,11 +17,26 @@
...
+	return NULL;
+}
+static int munmap(void *addr, size_t len) {
+	errno = ENOTSUP;
+	return -1;
+}
+#else
+#include <sys/mman.h>
+#endif
+
```
Akaricchi: Not a fan of this. Better just ifdef out all of `vfs_syspath_mmap` and `vfs_syspath_munmap`, and not even pass the pointers to `VFSSysPathNode` (they can just be NULL if mmap is not supported).

Creating an abstraction for file-mapping under `memory/` with all platform-specific implementations there (posix, win32, stub) would be even nicer, but it's a bit of work that's out of scope for this PR.

misc/ci/Dockerfile.switch
```diff
@@ -58,13 +58,17 @@ RUN apt-get install -y --no-install-recommends gh
...
 		switch-pkg-config && \
     dkp-pacman -S --needed --noconfirm `dkp-pacman -Slq dkp-libs | grep '^switch-'` && \
     dkp-pacman -Scc --noconfirm
+
+# compile SDL3 until changes are merged and packages are provided
+RUN git clone https://github.com/taisei-project/SDL.git -b taisei-3.4.12 --depth=1 /tmp/switch-sdl3 && \
+	cmake -S /tmp/switch-sdl3 -B /tmp/switch-sdl3/build \
+		-G Ninja -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE=/opt/devkitpro/cmake/Switch.cmake \
+		-DSDL_TEST_LIBRARY=OFF && \
+	ninja -C /tmp/switch-sdl3/build install
```
Akaricchi: Should just let it build as a subproject during the taisei build, like it's done for all of our releases. Saves us updating the docker image every time sdl needs to be updated/patched. Just needs a version bump in `subprojects/sdl3.wrap`

misc/ci/Dockerfile.switch
```diff
@@ -27,7 +27,7 @@ RUN apt update && \
 		apt-transport-https \
 		cmake
 
-RUN pip3 install \
+RUN pip3 install --break-system-packages \
 		meson==0.63.3 \
```
Akaricchi: Taisei needs meson >= 1.8.0

misc/ci/Dockerfile.switch
```diff
@@ -27,7 +27,7 @@ RUN apt update && \
...
 		ninja \
 		zstandard \
```
Akaricchi: Needs `backports.zstd` instead if Python <3.14, otherwise just remove this

misc/ci/Dockerfile.switch
```diff
@@ -27,7 +27,7 @@ RUN apt update && \
 		apt-transport-https \
 		cmake
 
-RUN pip3 install \
+RUN pip3 install --break-system-packages \
 		meson==0.63.3 \
 		ninja \
```
Akaricchi: Debian packages ninja, and all the python packages we need too, for that matter. Probably not an up to date meson though, unless we switch to forky.

Akaricchi (changes requested):

Akaricchi:
Will review the SDL stuff after i actually wake up

misc/ci/Dockerfile.switch
p-sam: Debian packages meson 1.0.1, still used pip to install it

misc/ci/Dockerfile.switch
p-sam: It's also now forcing usage of SDL3 subproject, should it stay like this ? Taisei SDL3 branch does not deviate from DKP's upstream and I plan to PR the pending changes

switch/crossfile.sh
```diff
@@ -46,4 +46,10 @@ system = 'nx'
 cpu_family = 'aarch64'
 cpu = 'cortex-a57'
 endian = 'little'
+
+[properties]
+cmake_toolchain_file = '$DEVKITPRO/cmake/Switch.cmake'
+
+[cmake]
+CMAKE_SYSTEM_NAME = 'NintendoSwitch'
```
Akaricchi, line 54: Is this actually needed? Meson will generate a cmake toolchain file by itself, and stuff in `cmake_toolchain_file` is appended to it.

meson.build
```diff
@@ -191,9 +191,14 @@ endif
 opt_gamemode = get_option('gamemode').require(host_machine.system() == 'linux',
     error_message : 'GameMode is Linux-specific')
 
+if host_machine.system() == 'nx'
+    dep_sdl3 = subproject('sdl3-cmake-wrapper').get_variable('sdl3_dep')
+else
+    dep_sdl3 = dependency('sdl3', version : '>=3.2.0', fallback : ['sdl3-cmake-wrapper', 'sdl3_dep'])
+endif
+
```
Akaricchi: This is not needed. It will fall back by default if the dependency is not installed (which should be the case in the CI environment), and you can force the fallback for a local build by passing `--force-fallback-for=sdl3` to `meson setup` (`meson configure -Dforce_fallback_for=sdl3` for an existing build dir).

I think it's fine to let it pick up the system sdl3 once the port is packaged.

switch/crossfile.sh
Akaricchi: Just read the relevant commit message. Is just overriding `CMAKE_SYSTEM_NAME` not enough here?

p-sam: the toolchain file along other things sets the `NINTENDO_SWITCH` flag that gates the switch port in the CMakeFiles, I would have preferred not including the CMAKE_SYSTEM_NAME directly here but it's defined like this: 

https://github.com/devkitPro/pacman-packages/blob/f103fe88e37180ecd0b7a9173b52b7580a54f71a/cmake/switch/Switch.cmake#L3

We could have the host_machine->system property match instead but it would not be lowercase like the other  meson standard values

Akaricchi: ok, seems reasonable then

Akaricchi:
Could you please rebase? There was a fix for the dockerfile pushed to master, but it was a misunderstanding, as your branch already contains the same fix. You can just override it with your changes.

Otherwise this seems good to go.

p-sam:
rebase done

PR #438, opened 2026-07-31
l10n: update zh-cn string

Swordsdeer:


po/zh_CN.po
```diff
@@ -5509,18 +5509,17 @@ msgid "Background rendering quality"
...
-#, fuzzy
 msgid "Background brightness"
-msgstr "背景渲染质量"
+msgstr "背景亮度"
 
 #: src/menu/options.c:845
 #, fuzzy
 msgid "Background contrast"
-msgstr "背景渲染质量"
+msgstr "背景对比度"
```
Akaricchi: Forgot to remove the fuzzy tag here

Akaricchi (changes requested):

PR #441, opened 2026-08-06
scripts/pack: honor SOURCE_DATE_EPOCH for zip entry timestamps

bmwiedemann:
`ZipInfo.from_file()` and `add_directory_entry()` both stamp zip entries with the mtime of the file on disk. For assets that ship with the source tree this is stable, but files produced during the build (e.g. compiled .mo translations) get the wall-clock build time instead, which differs between otherwise identical builds.

Clamp the embedded mtime to `SOURCE_DATE_EPOCH` when it is set, per https://reproducible-builds.org/specs/source-date-epoch/.

This patch was done while working on reproducible builds for openSUSE.

### DIFF
--- a/scripts/pack.py
+++ b/scripts/pack.py
@@ -2,6 +2,7 @@
 
 import dataclasses
 import hashlib
+import os
 import re
 import struct
 import zlib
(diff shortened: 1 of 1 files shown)


Akaricchi:
Looks good, thanks.

